Issue #73, opened 2018-06-14
Allow space in dataset mapping


MoritzFago:
As `zrepl/zfs/zfs.go:111` mentions, is space a valid char in a dataset name.

MoritzFago:
This error came back in 0.0.3 on freebsd.
With a mapping of 
```
  mapping: {
    "<":"foo bla"
  }
```
I get for a `zrepl test config` a 
```
error parsing config file: cannot parse job 'defiant': cannot automatically invert 'mapping' for prune job: mapping cannot be inverted: 'foo bla' is not a dataset path: %!s(MISSING): contains forbidden characters (any of '@#|        <>*')
```
I wasn't able to reproduce the error in a [playground](https://play.golang.org/p/ZIiqyMsdb-S).

PR #79, opened 2018-06-19
zfs: allow spaces in zfs names.

MoritzFago:
Space is a allowed character in zfs names accoring to
https://github.com/zfsonlinux/zfs/issues/439.
closes: #73.


problame:
Thanks!

PR #105, opened 2018-11-30
Permit peers to provide a cert chain (multiple certs). fixes #103

### LINKED ISSUE #103: Certificate trust with intermediate certificates fails (Wrong # of certs)
## Expectation: 
Peer certificates signed by an intermediary CA, containing the full certificate chain should be accepted by other peers that are configured with the root CA's certificate as their CA cert.

## Actual experience: 
Peer certificates with a full cert chain (host + intermediate) fail with the error: `unexpected number of certificates presented by TLS client`

Likely caused by code in `tlsconf.go`: 
```
if len(peerCerts) != 1 {
```

## Workaround: 
Adding the intermediate certificates to the trusted CA cert on the peers (and removing the full chain from the peer cert) causes the certificates to validate properly.

## Problem: 
Intermediate certs may be revoked at any time, and there may potentially be different intermediaries validating for a single root CA, so this isn't sustainable as a workaround.

## Further commentary:
It's pretty common practice when putting an internal CA in place to utilize intermediate CA servers, signed by the root ca, and to distribute only the root ca certificate for validation, such that as long as the client provides their certificate chain, it can be validated. At this time, `zrepl` appears to expect only one certificate to be sent by peers, preventing this from operating as expected. I'm sure that there is a reason this limitation was put in place, but for our (and likely other future users) needs this should be adjusted to support intermediary certificates. (Note that I believe some organizations may make use of multiple layers of intermediaries as well)

joshsouza:
I had some significant issues getting the build to work properly as documented, but after commenting out the `protoc` command in `generate` in the Makefile, it succeeded (I'm pretty sure that's NOT supposed to work) I was able to get this to build.
In testing with a client/peer cert that was a full chain (cert + intermediate) and a ca that was just the root ca (that signed the intermediate), I was able to successfully replicate.

I apologize, but I haven't taken the time to thoroughly look through the code here, and see if there are other repercussions of this change, and it might be better to eliminate this check entirely. However, I felt that if you put the code there to begin with, it probably has a purpose/need.

I would recommend properly reviewing what other repercussions this could have, and whether a better solution should be in place given a holistic view of `zrepl`, but this PR got us past our current blocker.


problame:
> had some significant issues getting the build to work properly as documented, but after commenting out the protoc command in generate in the Makefile, it succeeded (I'm pretty sure that's NOT supposed to work) I was able to get this to build.

For drive-by-contributions, this is what the CI is for ;)
Also, if you have docker installed, I'd recommend building inside docker as [documented here](https://zrepl.github.io/installation.html#compile-from-source).

The CI job was failing due to #106 , which I just fixed. Please rebase your commit onto `origin/master` and force push to this PR #105 ?

joshsouza:
Sorry for the delay, it's been a busy month with holidays and a bunch of other stuff. I've taken care of rebasing and will go verify that this version works as expected now.

joshsouza:
Confirmed that this is functional after building current version.

problame:
Thanks for your work!
No I'm only wondering whether it is correct to always check the 0th certificate if multiple are presented.

problame:
Nvmd, https://tools.ietf.org/html/rfc5246#section-7.4.2 defines that the peer cert must be the first in the list, so I guess we are ok. Should update the docs on this matter though.

joshsouza:
Would you like me to update the docs as part of this PR? (Just wanted to make sure that last comment was/wasn't an action item for me)

problame:
If you'd like to, sure! Should mention that the chain must be in the order in which verification happens, so peer cert first. See the RFC for food wording.

joshsouza:
Added documentation that hopefully makes sense to you and follows convention. Totally down to adjust to follow what you'd like to see there though.

problame:
I think that one sentence is totally sufficient!
I'll test this PR on my own setup and merge it before 0.1 is released.

That said, it'll be a few weeks until I can spend significant time on zrepl, so this PR will have to sit around open for a little longer.

joshsouza:
Not a problem. We're running on the version built with this patch for now without issue, so this PR isn't a blocker for us, just wanted to make sure to pay it forward.

PR #125, opened 2019-02-26
[WIP] Merge problame/replication_refactor into grpc_dataconn_rpc

problame:


replication/driver/replication_driver.go
```diff
@@ -0,0 +1,305 @@
...
+
+		run.l.Lock()
+		a1 := &attempt{
+			l:         l,
+			startedAt: time.Now(),
+			planner:   planner,
+		}
+		run.attempts = append(run.attempts, a1)
+		run.l.Unlock()
+		a1.do(ctx)
```
problame: Error handling is still missing here, given that we let gRPC fail early now:

* build a report (need to drop lock!)
* scan through it, find FSes with errors
* if errors, classify them, determine whether we should wait
* if we wait (non-permanent errors), make sure this is represented in the report
  * state? just a non-zero-valued 'sleep until' value?
* sleep
* build a new attempt
* copy FSes without errors (just pointers) + permanent (non-retryable) errors
* those that we want to retry: new, empty FSes
  * add a predecessor-pointer to the fs struct, need to pass that to the logic package...
* run driver as is, the fact that some FSes are pre-failed shouldn't be a problem
  * write tests for that

replication/driver/replication_driver.go
```diff
@@ -0,0 +1,305 @@
...
+		if block {
+			<-done
+		}
+		select {
+		case <-done:
+			return true
+		default:
+			return false
+		}
+
```
problame: newline

replication/driver/replication_driver.go
```diff
@@ -0,0 +1,305 @@
...
+	defer stepQueue.Start(1)()
+	var fssesDone sync.WaitGroup
+	for _, f := range a.fss {
+		fssesDone.Add(1)
+		go func(f *fs) {
+			defer fssesDone.Done()
+			f.do(ctx, stepQueue)
+		}(f)
+	}
+	// drop lock while waiting for wait group
```
problame: consider turning this into a feature of `chainmtx` ?

replication/driver/replication_driver_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (f *mockStep) TargetDate() time.Time {
+	return f.targetDate
+}
+
+func (f *mockStep) ReportInfo() *report.StepInfo {
+	return &report.StepInfo{From: f.ident, To: f.ident, BytesExpected: 100, BytesReplicated: 25}
+}
+
+func TestReplication(t *testing.T) {
```
problame: This kind of "test" is really flaky.

Since we properly separated the logic from the replication driver, it should be relatively straight forward to implement a recording proxy between driver and logic, and to capture test scenarios in a testbed setup, use those for regression testing.

Issue #126, opened 2019-03-02
pruner: filesystems with no correspondence on sink should not be pruned


JMoVS:
Hi,

in my tests, I have now run into a strange problem where zrepl first complained this:

```
2019-03-02T11:39:54+01:00 [ERRO][backup_system][pruning]: cannot get replication cursor prune_side="receiver" fs="zfsintern/Users" err="endpoint does not allow access to filesystem zfsintern/Users" invocation="1"
2019-03-02T11:39:54+01:00 [ERRO][backup_system][pruning]: entering error state after error invocation="1" prune_side="receiver" err="endpoint does not allow access to filesystem zfsintern/Users" state="ErrPerm"
```

this was the config at the time:

```
jobs:
  - type: sink
    name: "local_sink"
    root_fs: "zfssnaps/encRoot"
    serve:
      type: local
      listener_name: localsink

  - type: push
    name: "backup_system"
    connect:
      type: local
      listener_name: localsink
      client_identity: SergeantPepperSenior
    filesystems: {
      "zfsintern/Users/justin/Cloudy<": true,
      "zweischock/temp<": true
    }
    snapshotting:
      type: periodic
      interval: 10m
      prefix: zrepl_
    pruning:
      keep_sender:
      - type: last_n
        count: 10
      keep_receiver:
      - type: grid
        grid: 1x1h(keep=all) | 24x1h(keep=1) | 35x1d(keep=1) | 6x30d(keep=1) | 24x30d(keep=1) | 6x90d(keep=1)
        regex: "zrepl_.*"

```

and this is the output of zfs list

```
zfssnaps                                                                             1,03T   476G  2,55M  none
zfssnaps/encRoot                                                                     41,7G   476G   152K  none
zfssnaps/encRoot/SergeantPepperSenior                                                41,7G   476G   152K  none
zfssnaps/encRoot/SergeantPepperSenior/zfsintern                                      36,8G   476G   152K  none
zfssnaps/encRoot/SergeantPepperSenior/zfsintern/Users                                4,10G   476G   184K  none
zfssnaps/encRoot/SergeantPepperSenior/zfsintern/Users/justin                         4,10G   476G   184K  none
zfssnaps/encRoot/SergeantPepperSenior/zfsintern/Users/justin/Cloudy                  4,10G   476G   101M  none
zfssnaps/encRoot/SergeantPepperSenior/zfsintern/Users/justin/Cloudy/dataset1  1,06G   476G  1,03G  none
[...]

PR #111, opened 2018-12-08
rpc rewrite: control RPCs via gRPC, data transfer via own mechanism

problame:

PR #139, opened 2019-03-13
address new rpc in package replication & pruner

### LINKED ISSUE #123: pruner exits on filesystem-local errors
Original report in https://github.com/zrepl/zrepl/issues/70#issuecomment-466814052

> As of 0.1.0-rc3, in a pull-job, **the whole pruning phase fails** (for the remaining filesystems as well!), because zrepl can not get the replication cursor bookmark for the deleted filesystem. Left me with 25000 unpruned snapshots, until I found out.
> 
> In my specific case, I'm replicating my Proxmox VMs hard disks every 10 minutes. Each VM hard disk is a zvol and therefore a separate dataset. Removing a VM and it's hard disk in Proxmox eliminates the dataset, including snapshots and bookmarks. Consequently, zrepl runs into trouble, because it can't find the cursor bookmark anymore - and aborts the whole pruning process with a PermErr.
> 
> In my case, I would expect the existing snapshots to remain on the receiver side and to be pruned according to the pruning sieves. That would leave me with some backups until the period defined in the last pruning sieve.

I think I spotted a similar issue on my test setup: https://github.com/zrepl/zrepl/pull/111#issuecomment-458074582

problame:
This refactoring is mandated and enabled by #111 
Development happened in #125 

## Overview

* replication
  * separate the following 2 concerns:
    * doing the actual replication steps (package replication/logic)
    * driving the application, handling errors, responding to report requests from the CLI
* pruner
  * drop broken retry mechanism (grpc errors are classified incorrectly)
  - [x] drive-by fix #123  and #126 
  * omit a new retry mechanism for now, not as important as getting 0.1 out
* churn
  - [x] zrepl status improvements
  - [x] various rpc fixes
    - [x] goroutine leak fixed
  - [x] Travis config refactorings

## Todo

- [x] Adjust README.md project structure (also wrt #111 )
- [ ] Adjust user docs screen recording => asciinema ? -> postponed to #149 
- [x] mention automatic retries in features overview
- [x] get rid of `util/watchdog`


problame:
@jbliesener would you mind trying this branch against the problems described in [this comment]( https://github.com/zrepl/zrepl/issues/126#issuecomment-468912485)?

jbliesener:
Big patch. OK, I'll give it a try when the travis job terminates. Will there be any x86_64 binary that I can run directly or will I have to build?

problame:
> Big patch. OK, I'll give it a try when the travis job terminates. Will there be any x86_64 binary that I can run directly or will I have to build?

See link above (edited).

JMoVS:
I will try to do some stuff on the weekend. Thx for following up!

Debilski:
Hi, I was testing the above snapshot last week (freebsd) and it indeed solves a lot of problems that we had with a previous version (no more checksum errors after a few GB). But after a manual `zrepl signal wakeup` of a push job, I saw dozens (?) of `zfs send` jobs showing up in `htop` at the same time driving up the load and eventually crashing the server. Is this a deliberate behaviour that it starts so many sends at once?

Issue #131, opened 2019-03-04
Datasets with spaces in name aren't handled properly


JMoVS:
On Mac OS, some folders by default contain spaces, such as` "/Users/$user/Library/Application Support/"` - zrepl currently has a reproducible problem when encountering and trying to actually send such a filesystem.

```
2019-03-04T17:41:45+01:00 [ERRO][backup_system][repl]: dry run send request failed invocation="1" err="could not parse zfs send -n output: no match for info line (regex ^(full|incremental)(\\t(\\S+))?\\t(\\S+)\\t([0-9]+)$)"
2019-03-04T17:41:45+01:00 [ERRO][backup_system][repl]: error computing size estimate filesystem="zfsintern/Users/justin/Library/Application Support" invocation="1" err="could not parse zfs send -n output: no match for info line (regex ^(full|incremental)(\\t(\\S+))?\\t(\\S+)\\t([0-9]+)$)"
2019-03-04T17:41:45+01:00 [ERRO][backup_system][repl]: encountered global error while planning replication err="permanent global error: could not parse zfs send -n output: no match for info line (regex ^(full|incremental)(\\t(\\S+))?\\t(\\S+)\\t([0-9]+)$)" invocation="1"
```

JMoVS:
I guess that's an issue with the regex not properly ignoring spaces, right?

problame:
> On Mac OS, some folders by default contain spaces, such as "/Users/$user/Library/Application Support/" - zrepl currently has a reproducible problem when encountering and trying to actually send such a filesystem.

*Directory names* (paths, mountpoints) are not considered anywhere in the code.
This PR fixed handling of spaces in *dataset names*: https://github.com/zrepl/zrepl/pull/79

Would you please provide `zfs list` output of the sending side (where above log messages occurred)?

JMoVS:
> _Directory names_ (paths, mountpoints) are not considered anywhere in the code.
That's true and was inprecise on my end - I usually duplicate the underlying directory names in my dataset naming structure.

> This PR fixed handling of spaces in _dataset names_: #79
> 
> Would you please provide `zfs list` output of the sending side (where above log messages occurred)?


```
zfsintern                                                                                            140G  13,5G  1,48M  none
zfsintern/Users                                                                                      115G  13,5G   152K  none
zfsintern/Users/justin                                                                               115G  13,5G   152K  none
[...]

PR #151, opened 2019-03-21
fix handling of spaces in dataset names

problame:
addresses problems reported in #131 

@JMoVS  please report back whether these changes fixed your problems


JMoVS:
is that error not reproducible for you?

problame:
It was reproducible and I think I fixed it in the commit. If you don't want to test, I'll merge it right away ;)

JMoVS:
I'm a little short on time and have since taken out the spaces out of the datasets - if you were able to reproduce and now it's fixed, I give it a go :D

PR #200, opened 2019-07-17
fix #184 :  i386 build failure timeoutconn iovec len

### LINKED ISSUE #184: Build fails on FreeBSD i386 - rpc/dataconn/timeoutconn/timeoutconn.go
zrepl version: 0.1.1
OS: FreeBSD 11.2 or 12.0
Architecture: i386 only
Note: this build error is not seen on amd64 architectures

Excerpt of build error:
```
# github.com/zrepl/zrepl/rpc/dataconn/timeoutconn
rpc/dataconn/timeoutconn/timeoutconn.go:167:4: cannot use uint64(len(buffers[i])) (type uint64) as type uint32 in field value
rpc/dataconn/timeoutconn/timeoutconn.go:272:22: cannot use uint64(curVecNewLength) (type uint64) as type uint32 in assignment
```

Full build log error:
http://beefy10.nyi.freebsd.org/data/112i386-default/503555/logs/zrepl-0.1.1.log

Links to problematic code:
https://github.com/zrepl/zrepl/blob/master/rpc/dataconn/timeoutconn/timeoutconn.go#L167
https://github.com/zrepl/zrepl/blob/master/rpc/dataconn/timeoutconn/timeoutconn.go#L272

problame:
An actual end-to-end test on i386 would be nice before merging this.
On the other hand, this work merely makes zrepl compile on i386, we don't 'officially' support i386 because the Makefile doesn't produce a binary.

What do you think, @woodsb02?

### DIFF
--- a/rpc/dataconn/timeoutconn/timeoutconn.go
+++ b/rpc/dataconn/timeoutconn/timeoutconn.go
@@ -244,6 +244,14 @@ func (c Conn) readv(rawConn syscall.RawConn, iovecs []syscall.Iovec) (n int64, e
 func (c Conn) doOneReadv(rawConn syscall.RawConn, iovecs *[]syscall.Iovec) (n int64, err error) {
 	rawReadErr := rawConn.Read(func(fd uintptr) (done bool) {
 		// iovecs, n and err must not be shadowed!
+
+		// NOTE: unsafe.Pointer safety rules
+		// 		https://tip.golang.org/pkg/unsafe/#Pointer
+		//
+		//		(4) Conversion of a Pointer to a uintptr when calling syscall.Syscall.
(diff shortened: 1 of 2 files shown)


woodsb02:
Hi @problame, I’m not sure what you mean when you say the Makefile doesn’t produce a binary? Do you mean that i386 isn’t supported because the current build fails?

Does this actually need old hardware for end-to-end testing, or is it enough to install the i386 version of FreeBSD on new hardware or a VM and test?

problame:
It means: noone did any testing on i386, hence we don't produce a binary in the official Makefile.
No real hardware is needed, but I don't have the time to conduct a one-time manual integration test between two i386 VMs.

PR #203, opened 2019-07-27
Pre- and post-snapshot hooks

### LINKED ISSUE #74: pre- and post snapshot hooks

overhacked:
Jobs with periodic snapshots can run an executable before and/or after taking the snapshot,
configured in an optional `hooks` section of the `snapshotting` configuration.

Script output is passed through to zrepl logging. Scripts have a default 30 second timeout which can be configured. See docs/configuration/jobs.rst

Closes #74.


problame:
Thanks for your contribution!
I just reconfigured CircleCI to allow external builds, but I need to push a change to the `.circleci/config.yml` to make it work (secret sharing is disabled).
Would you please allow maintainer edits to your branch?
https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,74 @@ Note that you will have to trigger replication manually using the ``zrepl signal
        type: manual
      ...
 
+.. _job-snapshotting-hooks:
+
+Pre- and post-snapshot Hooks
+----------------------------
+
+Jobs with `periodic snapshots <job-snapshotting-spec_>`_ can run an executable before and/or after taking the snapshot,
+configured in an optional ``hooks`` section of the ``snapshotting`` configuration:
```
problame: Incorrect. `Periodic snapshotting can call optional pre- and post-hooks` (manual can't)

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,74 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+        prefix: zrepl_
+        interval: 10m
+        hooks:
+          pre: scripts/pre-snapshot.sh
+          post: /path/to/scripts/zrepl-notify.sh
+          timeout: 30s
+          keep_inconsistent: false
+      ...
+
+The path to the hook executables must either be absolute (e.g. ``/opt/bin/zrepl-notify.sh``) or is relative to the current configuration
```
problame: s/or is/or be/
s/current//

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,74 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+file directory (e.g. ``scripts/pre-snapshot.sh`` will be found in ``/etc/zrepl/scripts/pre-snapshot.sh``).
```
problame: s/scripts/hooks/g

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,74 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+The default is 30 seconds and may be specified in any units understood by `time.ParseDuration <https://golang.org/pkg/time/#ParseDuration>`_.
+
+The ``keep_inconsistent`` parameter specifies whether a snapshot will be taken after a failed pre-snapshot hook. The default
+is ``false``, meaning the snapshot will be skipped in the pre-snapshot hook process exits with an error or the timeout is exceeded.
+Setting ``keep_inconsistent: true`` will allow snapshots to be taken even if the pre-snapshot hook fails, but could result in
+inconsistent snapshots if, for example, the pre-snapshot script is asking a database to write its state to disk.
+
+Zrepl sets a number of environment variables for the hook processes:
+
+* ``ZREPL_HOOKTYPE``: either "pre" or "post"
```
problame: move this section before paragraph about 'no arguments can be specified'

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,74 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+
+The path to the hook executables must either be absolute (e.g. ``/opt/bin/zrepl-notify.sh``) or is relative to the current configuration
+file directory (e.g. ``scripts/pre-snapshot.sh`` will be found in ``/etc/zrepl/scripts/pre-snapshot.sh``).
+No arguments may be specified; create a wrapper script if zrepl must call an executable that requires arguments.
+The configuration may specify a ``pre`` script only, a ``post`` script only, or both.
+
+The ``timeout`` parameter is optional and specifies a period after which zrepl will kill the hook process and report an error.
+The default is 30 seconds and may be specified in any units understood by `time.ParseDuration <https://golang.org/pkg/time/#ParseDuration>`_.
+
+The ``keep_inconsistent`` parameter specifies whether a snapshot will be taken after a failed pre-snapshot hook. The default
```
problame: `keep_` is a rather overloaded name already (keep rules for pruning)
Maybe something like this?:  `snap_on_err: true|false`
Or `snap: always|no_error`?
Or `hook_error: ignore | nosnap`.

I'm not sure about the default of skipping the snapshot:
If users monitor zrepl log output (either through log monitoring or the prometheus metrics), or if they monitor zrepl as a black box (i.e. 'are new snapshots being replicated, check using own script'), skipping snapshots is fine to me.
If they don't monitor at all, I guess an inconsistent DB snapshot might still be better than none at all, which is an argument against skipping by default.
Any ideas?

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,74 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+      snapshotting:
+        type: periodic
+        prefix: zrepl_
+        interval: 10m
+        hooks:
+          pre: scripts/pre-snapshot.sh
+          post: /path/to/scripts/zrepl-notify.sh
+          timeout: 30s
+          keep_inconsistent: false
+      ...
```
problame: `hooks` is already plural.
Why not:

```
   hooks:
    - type: script
      pre: scripts/pre-snapshot.sh
      post: /path/to/scripts/zrepl-notify.sh
      timeout: 30s
      keep_inconsistent: false
```

This would leave a syntax-compatible option for future extension with multiple hooks in the future, e.g. built-in support for locking & unlocking mysql & postgres.

We could also just have `type`, but not a list:

```
   hook:
      type: script
      pre: scripts/pre-snapshot.sh
      post: /path/to/scripts/zrepl-notify.sh
      timeout: 30s
      keep_inconsistent: false
```

Opinions on multiple hooks / future extensibility?

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,125 @@
...
+	w.scanner.Scan()
+	if w.scanner.Err() == nil && w.scanner.Text() != "" {
+		w.logFunc(w.scanner.Text())
+	}
+
+	return nil
+}
+
+
+func RunHookCommand(ctx context.Context, commandDir string, command string, env map[string]string, timeout time.Duration) error {
```
problame: I would like to see this long argument list be converted into a struct => named arguments. Makes the call sites more expressive.

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,125 @@
...
+	w.scanner = bufio.NewScanner(&w.buf)
+	w.logFunc = logFunc
+	return w
+}
+
+func (w *logWriter) Write(in []byte) (int, error) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	bufCount, bufErr := w.buf.Write(in)
```
problame: s/bufCount/n/

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,125 @@
...
+	if bufErr != nil {
+		return bufCount, bufErr
+	}
+
+	if w.scanner.Scan() {
```
problame: should probably be a while loop, we are only going to log one line from the buffer otherwise, but the hook script could have logged several

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return bufCount, nil
+}
+
+func (w *logWriter) Close() error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	w.scanner.Scan()
```
problame: not draining buffer fully (could drop log lines). consider refactoring together with logWriter.Write

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,125 @@
...
+		return err
+	}
+
+	err = cmdExec.Wait()
+	if err != nil {
+		if cmdCtx.Err() == context.DeadlineExceeded {
+			l.Error(fmt.Sprintf("hook command timed out after %s", timeout))
+			return cmdCtx.Err()
+		}
+		l.WithError(err).Error("hook command exited with error")
```
problame: A nice bonus would be an aggregate of the hooks' stdout+stderr (say, capped to 1MB of memory). Because, in the event of hook failure, stdout output may also be valuable for users to debug their hook.

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,125 @@
...
+	scanner *bufio.Scanner
+	logFunc func(string)
+	level logger.Level
+}
+
+func NewLogWriter(mu *sync.Mutex, logFunc func(string)) *logWriter {
+	w := new(logWriter)
+	w.mu = mu
+	w.scanner = bufio.NewScanner(&w.buf)
+	w.logFunc = logFunc
```
problame: I would prefer the use of structured logging for this, i.e. a proper `github.com/zrepl/zrepl/logger.Logger`.
Called with `l.WithField("std{out|err}", line).Log(level, "hook output")`
Feel free to extend the `logger.Logger` interface with such a `Log(logger.Level, string)` method.

daemon/snapper/snapper.go
```diff
@@ -257,13 +262,58 @@ func snapshot(a args, u updater) state {
 			progress.state = SnapStarted
 		})
 
-		l.Debug("create snapshot")
-		err := zfs.ZFSSnapshot(fs, snapname, false)
-		if err != nil {
+		hookEnv := map[string]string{
+			"ZREPL_HOOKTYPE": "pre",
+			"ZREPL_PHASE": "Snapshotting",
```
problame: `ZREPL_PHASE` seems a bit arbitrary. What is the use-case you were thinking about? Future hooks for other functionality (pruning, replication)?

daemon/snapper/snapper.go
```diff
@@ -257,13 +262,58 @@ func snapshot(a args, u updater) state {
...
+			l.Debug("create snapshot")
+			err = zfs.ZFSSnapshot(fs, snapname, false)
+			if err != nil {
+				hadErr = true
+				l.WithError(err).Error("cannot create snapshot")
+			}
+			doneAt = time.Now()
+
+			if !hadErr {
+                hookEnv["ZREPL_HOOKTYPE"] = "post"
```
problame: ugh, not so pretty to mutate hookEnv here. factory function à la `func makeHookEnv(edge "pre"|"post")`?
Could also use it to generate the (proposed) argument struct to hooks.RunHookCommand.

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,74 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+* ``ZREPL_FS``: the ZFS filesystem name being snapshotted
+* ``ZREPL_SNAPNAME``: the zrepl-generated snapshot name (e.g. ``zrepl_20380119_031407_000``)
+
+As many pre-snapshot hook activities may affect only a single filesystem, the script or executable may wish
+to test the ``ZREPL_FS`` environment variable's contents. Example:
+
+::
+
+    #!/bin/sh
+
```
problame: `set -u` to catch reliance on undefined hook variables / typos, e.g. `ZREP_FS`  -> hook would exit with a proper error message.

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,74 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+    if [ "$ZREPL_FS" = "tank/data/postgresql" ]; then
```
problame: `case`? I would like to see a copy-pastable and little more elaborate example. Ideas:

* support both mysql and postgres. Other databases?
* derive the database name from `ZREPL_FS`

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,74 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+is ``false``, meaning the snapshot will be skipped in the pre-snapshot hook process exits with an error or the timeout is exceeded.
+Setting ``keep_inconsistent: true`` will allow snapshots to be taken even if the pre-snapshot hook fails, but could result in
+inconsistent snapshots if, for example, the pre-snapshot script is asking a database to write its state to disk.
+
+Zrepl sets a number of environment variables for the hook processes:
+
+* ``ZREPL_HOOKTYPE``: either "pre" or "post"
+* ``ZREPL_PHASE``: "Snapshotting"
+* ``ZREPL_FS``: the ZFS filesystem name being snapshotted
+* ``ZREPL_SNAPNAME``: the zrepl-generated snapshot name (e.g. ``zrepl_20380119_031407_000``)
```
problame: we should have `ZREPL_HOOK_DRYRUN=false|true` in here. There's no dry-run option ATM, but there could be in the future, and this is the only chance to get such a flag in without causing breakage in the future.
=> Also add this to the example hook.

overhacked:
Hmm, edits already allowed; I usually don’t turn that option off. I’m looking into the CI issues and I’ll run the make checks. Sorry for just throwing code over the wall. I may have been a bit excited to have gotten it working yesterday.

config/config.go
```diff
@@ -174,12 +175,32 @@ format: "human"
...
 type Global struct {
+	configDir string
+
 	Logging    *LoggingOutletEnumList `yaml:"logging,optional,fromdefaults"`
 	Monitoring []MonitoringEnum       `yaml:"monitoring,optional"`
 	Control    *GlobalControl         `yaml:"control,optional,fromdefaults"`
 	Serve      *GlobalServe           `yaml:"serve,optional,fromdefaults"`
 }
 
+func (g *Global) SetConfigDir(path string) error {
```
problame: Do you think always using full path names in the config hurts so much?
Relatiove to the config directory != working directory of the daemon => yet another relativ base path to keep track of...

daemon/snapper/snapper.go
overhacked: Exactly. I could imagine use cases for pruning and replication scripts, and it seemed like it would be fairly easy to implement if desired. I included the `ZREPL_PHASE` environment variable to help with that future expansion.

docs/configuration/jobs.rst
overhacked: But `manual` != `periodic`, or do I misunderstand some terminology? I meant to refer only to jobs with `snapshotting.type == "periodic"`.

docs/configuration/jobs.rst
overhacked: I think the list would make sense if each item in the list had a `filesystem` item that specified which filesystems it operated on. I actually like that idea because it avoids the whole "here's how to filter by filesystem in your script" tutorial.

I'm going to push all the other changes and then take a crack at using the filesystem filter code in this location.

daemon/hooks/hooks.go
overhacked: Hmm. Thinking about how to do this, because if `Write()` calls `Scan()` exhaustively, then the Scanner terminates prematurely and returns false forever. I'm thinking that I need to use something other than the Scanner here.

daemon/hooks/hooks.go
overhacked: Agreed, this needs to be a loop that drains the buffer fully and exhausts the scanner.

overhacked:
Pending some review items above, still to be resolved, I've pushed some major design changes. There can now be multiple hooks per job, and each hook has a `filesystems` filter, allowing **a lot** more flexibility. This change ended up refactoring a good bit of the hook running code into the daemon/hooks package, so that was a nice benefit. The logging is a bit cleaner, and I think that the `logWriter.Write()` method now fully drains the buffer every time. The buffer will be Reset() if it exceeds 1MB, but I think that is mostly unnecessary unless a hook process writes more than 1MB at a time. I haven't yet completed documenting the changes; mostly I need to work up a better example script.

overhacked:
One or two remaining review items above for me to work on today. I'll add another commit or two to address. Would you prefer that I rebase and force push before you review to have a cleaner history?

problame:
No rebases / force pushes please, Github is remarkably smart (it tracks the last reviewed commit and allows me to just inspect the changes you made).

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,233 @@
...
+					hadFatalErr = true
+					return
+				}
+			}
+		}
+	}
+	return
+}
+
+func RunHookCommand(ctx context.Context, command string, env HookEnv, timeout time.Duration) error {
```
problame: Would like to see this return a more specific hook error type:

```
type Error struct {
  Command string
  Args []string // currently always empty
  Env []string
  Edge HookEdge
  RunTime time.Duration
  WaitErr error
  CapturedStdoutStderrCombined []byte
}

func (e *Error) Error() string {
 // do what you do in ``fmt.Errorf` here  
}
```

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,233 @@
...
+	err := cmdExec.Start()
+	if err != nil {
+		l.WithError(err).Error("hook command failed to start")
+		return err
+	}
+
+	err = cmdExec.Wait()
+	if err != nil {
+		if cmdCtx.Err() == context.DeadlineExceeded {
+			l.Error(fmt.Sprintf("hook command timed out after %s", timeout))
```
problame: `&Error{WaitErr: context.DeadlineExceeded}`

config/config.go
```diff
@@ -361,6 +362,19 @@ type JobDebugSettings struct {
 	RPCLog bool `yaml:"rpc_log,optional,default=false"`
 }
 
+type HookList []HookSettings
+
+type HookSettings struct {
```
problame: Each hook type should be its own Go type. Look at `PruningEnum`:

https://github.com/zrepl/zrepl/blob/afed7627749bee3726a06f7855a11a82b6769b09/config/config.go#L271-L273

https://github.com/zrepl/zrepl/blob/afed7627749bee3726a06f7855a11a82b6769b09/config/config.go#L416-L424

 https://github.com/zrepl/zrepl/blob/afed7627749bee3726a06f7855a11a82b6769b09/config/config.go#L153-L155

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func getLogger(ctx context.Context) Logger {
+	if log, ok := ctx.Value(contextKeyLog).(Logger); ok {
+		return log
+	}
+	return logger.NewNullLogger()
+}
+
+const MAX_HOOK_LOG_SIZE int = 1048576
```
problame: `1 << 20` is more readable IMHO

daemon/snapper/snapper.go
```diff
@@ -46,6 +52,8 @@ type args struct {
 	interval       time.Duration
 	fsf            *filters.DatasetMapFilter
 	snapshotsTaken chan<- struct{}
+	hooks          config.HookList
+	dryRun         bool
```
problame: `dryRun` is dead (always false) => I wouldn't even add it at this point

daemon/snapper/snapper.go
```diff
@@ -241,6 +252,16 @@ func snapshot(a args, u updater) state {
 		plan = snapper.plan
 	})
 
+	// Prepare hook filters
+	hookRun, err := hooks.NewHookRunner(a.ctx, &a.hooks, "snap", a.dryRun)
+	if err != nil {
+		a.log.WithError(err).Error("error in configuration; could not create snapshot hook runner")
```
problame: error only happens when creating a new dataset filter atm, which if validated during config parse, actually never causes an error.
Either we do real checking in NewHookRunner, e.g. `does the binary we want to execute exist` or I'd say we shift the filter creation to the `Snapper` constructor from config, i.e., https://github.com/zrepl/zrepl/blob/5b97953bfb099d523c9158c533315162f54f4847/daemon/snapper/snapper.go#L118

daemon/snapper/snapper.go
```diff
@@ -36,7 +38,11 @@ type snapProgress struct {
 	doneAt time.Time
 
 	// SnapErr
-	err error
+	errs struct {
+		preHook  []error
```
problame: These should be the new `*snapper.Error` type I proposed (See `RunHookCommand`).
The piece of context that's missing to render `snapProgress` with `errs.preHook != nil` into a proper error message that explains to the user what hook failed why is the information **which** hook in the configuration produced the error.
Perhaps we should give names to the hooks in the configuration, and include the hook name in `snapper.Error`?

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,145 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+      elapsed=$(($elapsed + 1))
+    done
+
+    if [ $elapsed -ge $MYSQL_LOCK_TIMEOUT_SECONDS ]; then
+      2> echo "Couldn't acquire MySQL READ LOCK"
+      exit 1
+    fi
+
+.. code-block:: bash
+    :caption: mysql-post-snapshot.sh
```
problame: I think pre and post code belong into the same file.
Of course that's a matter of taste, but in particular, you have duplicate definitions of setup-specific information like `MYSQL_LOCK_FILE` in both files, which calls for trouble (admin will only change one file => pre and post use different locks => things hang until timeout => unhappy admin)

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,145 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+* ``ZREPL_SNAPNAME``: the zrepl-generated snapshot name (e.g. ``zrepl_20380119_031407_000``)
+* ``ZREPL_DRYRUN``: indicates whether a dry run is in progress so scripts can skip some snapshot tasks
+
+No arguments may be specified; create a wrapper script if zrepl must call an executable that requires arguments.
+Each hook configuration may specify a ``pre`` script only, a ``post`` script only, or both. Each script can include an optional ``filesystems`` filter, limiting which filesystems the hook runs for. This uses the same |filter-spec| as jobs.
+
+The ``timeout`` parameter is optional and specifies a period after which zrepl will kill the hook process and report an error.
+The default is 30 seconds and may be specified in any units understood by `time.ParseDuration <https://golang.org/pkg/time/#ParseDuration>`_.
+
+The ``stop_after_error`` parameter specifies whether the snapshot process will stop after a failed hook. The default
```
problame: I think you found an even better name for this flag in the implementation: `error_is_fatal`

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,145 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+          timeout: 30s
+          stop_after_error: false
+        - type: script
+          pre: /path/to/hooks/special-snapshot.sh
+          filesystems: {
+            "tank/special": true
+          }
+      ...
+
+The ``type: script`` parameter is mandatory. Future versions of Zrepl may support other hook types.
```
problame: s/Zrepl/zrepl`

daemon/snapper/snapper.go
```diff
@@ -257,21 +278,62 @@ func snapshot(a args, u updater) state {
...
+			doneAt = time.Now()
+
+			if !hadErr {
+				l.Debug("post-snapshot hook")
+				var postHookHadFatalErr bool
+				postHookErr, postHookHadFatalErr = hookRun.RunFilesystemHooks(fs, hooks.Post, hookEnvExtra)
+				for _, err := range postHookErr {
+					l.WithError(err).Error("cannot run post-snapshot hook(s)")
+					if postHookHadFatalErr {
+						l.Error("the post-hook run was aborted due to a fatal error in the previous hook")
```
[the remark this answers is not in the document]
overhacked: `hadErr` really could be renamed to `hadSnapErr`, because all the following logic takes actions that relate to retrying a snapshot (i.e. state `ErrorWait`). I did not set `hadErr` here, because it seemed that the only consequences of a failed post-hook with `ErrIsFatal == true` should be to skip following post-hooks and log the error.

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,233 @@
...
+func RunHookCommand(ctx context.Context, command string, env HookEnv, timeout time.Duration) error {
+	l := getLogger(ctx).WithField("command", command)
+
+	cmdCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	cmdExec := exec.CommandContext(cmdCtx, command)
+	cmdEnv := os.Environ()
+	for k, v := range env {
+		cmdEnv = append(cmdEnv, fmt.Sprintf("%s=%s", k, v))
```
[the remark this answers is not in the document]
overhacked: Quoting is bad here. The only POSIX requirement for the raw environment strings is that the key doesn't contain an `=`. I'm not checking for that here. I feel that, as these are not exposed to user input, keeping it simple is better.

Go respects the POSIX semantics: https://golang.org/src/syscall/env_unix.go#L85

daemon/snapper/snapper.go
```diff
@@ -257,21 +278,62 @@ func snapshot(a args, u updater) state {
...
-		if err != nil {
-			hadErr = true
-			l.WithError(err).Error("cannot create snapshot")
+		var doneAt time.Time
+
+		var snapErr error
+		var preHookErr []error
+		var postHookErr []error
+		hookEnvExtra := map[string]string{
+			"fs":       fs.ToString(),
```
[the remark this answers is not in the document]
overhacked: I left it separated, thinking of future hooks like a post-job hook where a single `fs` name wouldn't apply.

daemon/snapper/snapper.go
overhacked: So put in in the HookEnv as a `false` literal for now, per your earlier comment that you'd like it to be set in the hook process environment even if it's not currently used?

problame: Yup. I'm currently still undecided whether we want it to be

1. `false`, `true` => `if [ "$ZREPL_DRYRUN" = "false" ]; then X; else Y; fi` 
2. "ZREPL_DRYRUN=", "ZREPL_DRYRUN=true" => `if [ -z "$ZREPL_DRYRUN" ]; ...`

The second variant is a little more typical.

daemon/snapper/snapper.go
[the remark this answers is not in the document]
problame: > it seemed that the only consequences of a failed post-hook with ErrIsFatal == true should be to skip following post-hooks and log the error.

See the review-wide commend on Pre-only Pre-and-Post, and Post-Only hooks.

daemon/snapper/snapper.go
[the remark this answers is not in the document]
problame: Ah, I see:

https://github.com/zrepl/zrepl/pull/203/files#diff-dce3c966e41dcdc41961b89f87185a4cR117

I think the naked "fs" and later 'magic' conversion into `ZREPL_FS` makes it actually harder to understand where `ZREPL_FS` is set, e.g. when grepping the source code.
I would suggest extracting all ZREPL_ environment variables into constants `hooks.Env{Hooktype,Phase,DryRun,FS,...}` of the hooks package, including `ZREPL_FS`. Then just use `hooks.EnvFS` instead of the naked `fs` and get rid of the 'magic' logic.

problame:
@overhacked replied to the review comments.
Not sure if you saw my review-wide remarks on PRE-ONLY, PRE-AND-POST, and POST-ONLY snapshots. (sry, GitHub apparently doesn't allow linking to it...).

What do you think?

problame:
I merged FilteredHook and Hook into one. Feel free to grab this patch (was easier to write it directly than make code annotations ;) )

https://gist.github.com/problame/4fb420cfed89f68a4f56d405d527d90b

daemon/hooks/hook_command.go
```diff
@@ -0,0 +1,142 @@
...
+		runErr.Err = err
+		return
+	}
+	logErrWriter := NewLogWriter(&scanMutex, l, logger.Warn, "stderr")
+	logOutWriter := NewLogWriter(&scanMutex, l, logger.Info, "stdout")
+	defer logErrWriter.Close()
+	defer logOutWriter.Close()
+
+	cmdExec.Stderr = io.MultiWriter(logErrWriter, combinedOutput)
+	cmdExec.Stdout = io.MultiWriter(logOutWriter, combinedOutput)
```
problame: combinedOutput (*circbuf.Buffer) is written to by both the stderr and stdout copier goroutine, but circbuf is not goroutine safe => race condition.

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+//go:generate stringer -type=HookEdge
+type HookEdge uint
+
+const (
+	All HookEdge = 1 << iota
+	Pre
+	Post
+)
```
problame: This smells like a bitmap to me... 0x01 = Pre, 0x02 = Post

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (e HookEdge) RunsDuring(runningEdge HookEdge) bool {
+	if e == All {
```
problame: ... rendering this a bitmap test

daemon/hooks/hooks.go
```diff
@@ -0,0 +1,68 @@
...
+		return true
+	} else if e == runningEdge {
+		return true
+	}
+	return false
+}
+
+type HookList []FilteredHook
+
+type FilteredHook struct {
```
problame: see https://gist.github.com/problame/4fb420cfed89f68a4f56d405d527d90b

daemon/hooks/hook_logging.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func getLogger(ctx context.Context) Logger {
+	if log, ok := ctx.Value(contextKeyLog).(Logger); ok {
+		return log
+	}
+	return logger.NewNullLogger()
+}
+
+const MAX_HOOK_LOG_SIZE int = 1 << 20
```
problame: please use envconst

 https://github.com/zrepl/zrepl/blob/c87759affef880d18155a4776dd90a1183912917/util/envconst/envconst.go#L28 

If you need an int, adjust envconst as needed (copy pasta)

daemon/hooks/hook_runner.go
```diff
@@ -0,0 +1,70 @@
...
+	l := getLogger(r.context)
+
+	for _, hf := range *r.filters {
+		passFilesystem, err := hf.Filter.Filter(fs)
+		if err != nil {
+			retErr = append(retErr, RunError{Err: fmt.Errorf("unexpected filter error: %s", err)})
+			return
+		}
+		passEdge := hf.Hook.Edge().RunsDuring(edge)
+
```
problame: see https://gist.github.com/problame/4fb420cfed89f68a4f56d405d527d90b

daemon/hooks/hook_command.go
overhacked: I wasn't happy with circbuf.Buffer, as it required allocating too much memory up front, so I wrote a thread-safe `CircularLog` class that resizes the internal buffer as needed.

daemon/snapper/snapper.go
overhacked: I support the latter, as it's more shell-like.

daemon/hooks/hook_command.go
problame: If you are implementing a custom log buffer anyways, maybe try to copy the overflow handling from https://godoc.org/os/exec#ExitError 

>  // If the error output is long, Stderr may contain only a prefix
    // and suffix of the output, with the middle replaced with
    // text about the number of omitted bytes.

Not a hard requirement, however.
If possible, try to make it a sub-package of `./utils`, could imagine using it in package `./zfs` at some point.

overhacked:
I think that I've addressed all your comments in the above push (9 commits). The implementation of pre-and-post/pre/post/cleanup hooks is a little different than the stack mechanism you proposed, but it results in the same ordering without having to manage a queue.

Looking forward to your feedback!

overhacked:
I took a look at the CI errors, and they are mostly `make lint` errors in files I didn't change. There is one linter error in <https://github.com/zrepl/zrepl/pull/203/commits/809ff78f4955b0dff38d7dbc7b3403d84f6fb1c5#diff-7c76a8902f7ea127993004d20a333386R21-R27>, but it is complaining about using implicit type expressions in a `const(…)` statement, which I see elsewhere in the code.

overhacked:
FYI, working on some functional tests for hooks.

overhacked:
Thanks for the detailed feedback! I appreciate you taking the time to push me to make this an even better PR. A couple questions about the hook driver initiating the snapshot:
1. Could the function pointer it calls be a closure of this code from `snapper.go`: https://github.com/zrepl/zrepl/blob/c4be60c29fbe13d780506b66a3da1bc993155d4b/daemon/snapper/snapper.go#L254-L275
2. If implementing the “snapshot batching” functionality, would that change the hook driver? Couldn't its snapshot callback function just then call a queuing method of Snapper that handled the periodic flushing of the queue to a channel program?

config/config.go
```diff
@@ -501,6 +521,15 @@ func (t *SyslogFacility) UnmarshalYAML(u func(interface{}, bool) error) (err err
 	return nil
 }
 
+func (t *HookEnum) UnmarshalYAML(u func(interface{}, bool) error) (err error) {
+	t.Ret, err = enumUnmarshal(u, map[string]interface{}{
+		"pre-and-post-command": &HookCommand{},
```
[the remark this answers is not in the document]
overhacked: I thought about this before giving it such a long name. I worry that if we give it a short, intuitive name like `command` that there will be sysadmin mistakes of putting a pre-hook in as a `command`. I've never used a hook system before in which a single command was run as both pre and post with only a change in environment. I like the elegance of the design you've suggested, but I'm concerned that sysadmins who don't read the docs carefully enough are likely to accidentally run their pre-hooks as post-hooks also if the name of the type isn't clear.

I agree it could be less clunky. Maybe `multi-command`? `all-command` sounds nice, but what if there are different edges in the future? (… thinking …) Maybe the right meaning here is expressed by `custom-command`, which hints to the administrator that she has to include some logic in the hook process for things to work properly. That also leaves the door open to edges other than `pre` and `post` in the future. All *custom* scripts should be implemented with the `if pre; elif post; else unsupported error` pattern so they break in an understandable way if applied to a new version with new edge semantics.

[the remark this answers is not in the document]
problame: I like the approach that the dehydrated Let's Encrypt client is taking:
https://github.com/lukas2511/dehydrated/blob/master/CHANGELOG#L53

Also Ctrl-F hook through the changelog, and have a look at the example hook https://github.com/lukas2511/dehydrated/blob/master/docs/examples/hook.sh

Bottom line: one hook.sh for all kinds of events. People seem to deal with it fairly well given the example hook as a starting point.

problame:
>     1. Could the function pointer it calls be a closure of this code from `snapper.go`: https://github.com/zrepl/zrepl/blob/c4be60c29fbe13d780506b66a3da1bc993155d4b/daemon/snapper/snapper.go#L254-L275

Yes, that' spretty much the idea.
 
>     2. If implementing the “snapshot batching” functionality, would that change the hook driver? Couldn't its snapshot callback function just then call a queuing method of Snapper that handled the periodic flushing of the queue to a channel program?

Pretty much that. The queuing method would need to block - think of it as a barrier that queues up 'take-snapshot' requests for about 500ms, then takes all snapshots that have been requested so far, then unblocks all requesters, allowing them to run their post hooks in parallel.
I.e., the only requirement to the hook driver is that multiple instances of it can run in parallel (one per FS).

daemon/hooks/hook_circlog.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		copy(newBuf, cl.buf[:cl.writeCursor])
+		cl.buf = newBuf
+		cl.size = newSize
+	}
+
+	// If data to be written is larger than the max size,
+	// discard all but the last cl.size bytes
+	if n > cl.size {
+		data = data[n-cl.size:]
```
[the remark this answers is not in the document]
overhacked: Implemented this, but didn't use a `bool` flag, because it wouldn't be concurrency-safe. Another write could sneak in before `Bytes()` is called, and the buffer would no longer be overrun. It would just have wrapped as designed.

Instead of flagging the overrun, I've just altered the first few bytes of the truncated write before I put it into the buffer.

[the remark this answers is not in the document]
problame: Why doesn't `Bytes()` also take the mtx?

overhacked:
If you don’t object, I’ll leave the snapshot batching/callback for another PR that I’m happy to contribute to. I’d make sure that the hook planner is as ready as it can be to accommodate that functionality.

problame:
> If you don’t object, I’ll leave the snapshot batching/callback for another PR that I’m happy to contribute to. I’d make sure that the hook planner is as ready as it can be to accommodate that functionality.

Yes, converting things to  a driver/planner approach is totally sufficient!

overhacked:
Just pushed most of the requested changes above. One pending item is `config` package unit testing for default values of `config.HookCommand`. I'd like to give the documentation another look-over before I call it good, but I'd appreciate your editorial eye on it, too.

I put significant effort into printing a human-readable string of the hook plan for testing, and I'm wondering if there is a place that that string should be put in the status updates?

Also, the unit tests for `util/circlog` have 97.1% coverage, not 100%, because there are two `panic()` statements in a private helper function that are unreachable. Basically, the code of the public API of `CircularLog` will never call the helper in such a way that it panics, but I did not want to remove those two `panic()`s for the sake of future maintainers.

daemon/hooks/hook_plan.go
```diff
@@ -0,0 +1,249 @@
...
+	l := getLogger(ctx)
+
+	plan := &HookJobPlan{
+		context: ctx,
+		phase:   phase,
+		env:     extra,
+	}
+
+	for _, h := range *hooks {
+		//l := l.WithField("hook", h)
```
problame: remove this

daemon/hooks/hook_plan.go
```diff
@@ -0,0 +1,249 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unexpected filter error: %s", err)
+		}
+		if passFilesystem {
+			/*
+			   Iterate over each edge in the bitfield,
+			   creating multiple entries in the plan if
+			   the hook has multiple edges.
+			*/
+			//l.Debug(fmt.Sprintf("adding hook to edge %s", edge.String()))
```
problame: or leave it in and enable this debug stmt, if you think it will need / aid debugging.
consider introducing a `func debug` enabled by an env variable as found in other packages (e.g. rpc)

daemon/hooks/hook_plan.go
```diff
@@ -0,0 +1,249 @@
...
+	// Use user-provided order as secondary key, so
+	// this is a stable sort.
+	edge_i := r.steps[i].Edge
+	edge_j := r.steps[j].Edge
+	if edge_i.Before(edge_j) {
+		return true
+	} else if edge_j.Before(edge_i) {
+		return false
+	}
+	return i < j
```
problame: I'm not sure that sort.Stable guarantees that i, j are in user-provided order.
Probably not, unless the sorting algorithm plans everything ahead of time.
=> add `configUserOrder int` member to `step` that reflects the index of the hook specified in the configuration file?

daemon/hooks/hook_plan.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func (r *HookJobPlan) Swap(i, j int) {
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	r.steps[i], r.steps[j] = r.steps[j], r.steps[i]
+	r.steps[i].Order = i + 1
+	r.steps[j].Order = j + 1
```
problame: what is the purpose of the `Order` member? Why do we change it while sorting? 
(read comment in `func Less` above first)

daemon/hooks/hook_plan.go
```diff
@@ -0,0 +1,249 @@
...
+				r.mtx.Lock()
+				e.Hook.Disable()
+				r.mtx.Unlock()
+				r.mtx.RLock()
+			}
+			// skip to Cleanup hooks if err_is_fatal == true
+			if e.ErrIsFatal {
+				hadFatalErr = true
+				allowedEdges = Cleanup
+				l.Error("the hook run was aborted due to a fatal error in the previous hook")
```
problame: it's not the previous hook, it's *this* hook => s/previous/this/

daemon/hooks/hooks.go
```diff
@@ -13,25 +13,55 @@ type Hook interface {
...
+	Disabled = HookEdge(0)
+	All      = HookEdge(^uint(0)) // max value, all bits == 1
 )
 
 func (e HookEdge) RunsDuring(runningEdge HookEdge) bool {
 	return e&runningEdge != 0
 }
 
+func (e HookEdge) StringForPhase(phase HookPhase) string {
+	// Cleanup is a special type of Post-hook,
```
problame: `special`: someone is going to scratch their head about this in two years. Refer to the place where you explain how config semantics are translated to the edge semantics. If such a section doesn't exist, introduce one in a `hook_docs.go` package-level comment or where the Edge constants are defined.

util/circlog/circlog_test.go
```diff
@@ -0,0 +1,163 @@
...
+		pattern := []byte{0xDE, 0xAD, 0xBE, 0xEF}
+		writeBytes := bytes.Repeat(pattern, size/4)
+		if len(writeBytes) < size {
+			writeBytes = append(writeBytes, pattern[:size-len(writeBytes)]...)
+		}
+		n, err := w.Write(writeBytes)
+		if err != nil {
+			return n, err
+		}
+		/*
```
problame: why is this in here?

daemon/hooks/hook_plan.go
overhacked: When I implemented this I had a golang-specific reference that showed this final comparison as the way to implement Stable / original order sorting as a second key. I've since lost the URL for that original source, but I've carefully read through [Go's implementation of Stable()](https://golang.org/src/sort/sort.go?s=10596:10623#L346). Given that `i` and `j` are indices into the slice, which we construct in config order, the value of `i < j` therefore `Less(i, j)` will always reflect user-provided order.

daemon/hooks/hook_plan.go
overhacked: `Order` is there only for `HookJobPlan.String()`, so it can print the step number of dependencies. The offset within the `steps` slice can't be gotten at via `hookPlanEntry.Dependencies`, so we have to store the Order explicitly and keep it updated. Either that or iterate completely through `steps` for every dependency to look up the offset.

util/circlog/circlog_test.go
overhacked: Because the `writeFixedSize` closure wraps the `io.Writer` interface that is passed to it, so it returns the same values that `Writer.Write()` would. I could have put `require.NoError()` inside the closure, but I prefer to keep all the assertions farther down in the test body, where I call `writeFixedSize()`.

problame: I was referring to the commented-out code block...

daemon/hooks/hook_plan.go
problame: So `Less` is only correct / deterministic for `sort.Stable` then.

util/circlog/circlog_test.go
overhacked: I must have been tired when reviewing your comments. Of course, that's some dead code that needs to be taken out. Fixed.

daemon/hooks/hook_plan.go
overhacked: Agreed. Added `configUserOrder`.

daemon/hooks/hook_plan.go
overhacked: Rewrote the dependency planner so that all sorting and dependency resolution is within `hook_plan.go`. Order is now an integral part of the Dependencies used during `Run()`

daemon/hooks/hooks.go
overhacked: With the dependency planner rewrite, `Cleanup` is now gone.
### Before
* `Hook.Edge() & Post > 0 && Hook.AlwaysRun() == true`
  ≡
  `hookPlanEntry.Edge == Cleanup`
* Sort by
  1. `hookPlanEntry.Edge`
  2. `HookJobPlan.steps` slice order
### After
* No `Cleanup` edge
* Sort by
  1. `hookPlanEntry.Edge`
  2. `Hook.AlwaysRun() == true` > `Hook.AlwaysRun() == false`
  3. `hookPlanEntry.configOrder`

daemon/hooks/hook_plan.go
```diff
@@ -0,0 +1,249 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/zrepl/zrepl/zfs"
+)
+
+type HookReport struct {
+	Command                      string
+	Args                         []string // currently always empty
```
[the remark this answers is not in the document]
overhacked: So that `HookReport.Error()` is correct for future hook types (like database checkpoint hooks) that *do* have arguments.

overhacked:
I took a look at the CI errors, but I'm not sure I yet understand what's going on. It looks more like tooling or configuration problems than code issues?

problame:
* [x] Refactoring to stack-based execution model cherry-picked by @overhacked 
* [x] @problame's patches since that refactoring cherry-picked by @problame 
* [x] Squashed everything into bffc0463b7020544491f65f2ba36d0001e971c23
* [ ] Rebase onto master and adopt Go modules (@problame ready)
* [ ] Fix ugly callback log message leaking function name
* [ ] Check source code for
  * [ ] FIXME's and TODO's
  * [ ] dead code
  * [ ] missing test coverage

daemon/hooks/hooks_docs.go
```diff
@@ -0,0 +1,24 @@
...
+ The hook list can be optionally filtered by filesystem, if `FilterHookListForFilesystem(hooks, fs)` is called by the job when it runs (i.e. during `snapper.snap()`). If `FilterHookListForFilesystem()` is not called,
+ any `filesystems` filters specified in the YAML will be ignored.
+
+ The job must create a `CallbackHook` by passing a filesystem and a closure, taking no arguments and returning a nillable error, to `NewCallbackHookForFilesystem()`. Future implementations of jobs with non-filesystem-scoped hooks will need to add a `NewCallbackHook()` method that creates a `CallbackHook` matching all filesystems/without a filter.
+
+ The job's hook list, filtered or not, must then be passed to `NewHookJobPlan()` along with the `CallbackHook`.
+
+ Once created, the plan is invoked with `HookJobPlan.Run(context.Context, dryRun bool)`. The context passed should contain a logger (provided in `daemon/logging/build_logging.go` by `hooks.WithLogger()`). If `dryRun == true`, the environment variable `ZREPL_DRYRUN` is set to `"true"` for any hook command processes. Future hook types may use `dryRun` to provide other diagnostic output.
+*/
+package hooks
```
problame: above docs are out of date

docs/configuration/jobs.rst
```diff
@@ -173,6 +173,230 @@ Note that you will have to trigger replication manually using the ``zrepl signal
...
+    $DRYRUN /usr/local/bin/psql \
+      --echo-errors \
+      --no-password \
+      --no-psqlrc \
+      -U postgres \
+      -c "SET statement_timeout TO $PSQL_TIMEOUT; CHECKPOINT;"
+    # Exit with psql's exit code
+    exit $?
+
+**MySQL** ``LOCK_TABLES`` **During Snapshot**
```
problame: move those scripts to an example file, probably replace them with a template for people to copy instead

(we have built-in support for mysql and postgres, no need to show what these builtins do in such detail in the docs)

overhacked:
A few issues I've spotted that have shown up after you've rebased:


1. The bump from tcell v1.0.0 to v1.2.0 introduced a dependency on the `infocmp` binary from ncurses, which is not installed by default on all platforms. The effect is that I was unable to run `zrepl status` without installing the ncurses package. Not too big a deal, but worth noting in the documentation. It affects FreeBSD at least.


2. In the new Snapshotting status display, errors seem to never disappear. The "check logs for details" message stays even after the next successful snapshot run.


3. When a hook error occurs, the duration of the snapshot is reported as `-2562047h47m16.854775808s` (or similar), which seems to be an uninitialized integer?


Sorry to not have any solutions yet to offer. I'll be more productive when I get back from this conference on Monday.

problame:
Will look into these problems, tcell is in #204 so we'll continue debugging this there (currently installing FreeNAS in a VM to determine how well zrepl works out of the box there).

problame:
Allright, it's merged! Thanks @overhacked for all the time you spent on this!

PR #226, opened 2019-10-06
Fix invalid comment syntax

juergenhoetzel:
`systemd` can not parse the `ProtectHome` setting:
```
journalctl -u zrepl -p err
-- Logs begin at Sat 2019-08-24 03:04:22 CEST, end at Sun 2019-10-06 13:21:35 CEST. --
Okt 05 19:30:01 pidsley systemd[1]: /etc/systemd/system/zrepl.service:21: Failed to parse protect home value, ignoring: read-only   # DEBIAN STRETCH

sudo systemctl show zrepl|grep ^ProtectHome
ProtectHome=no
```

I just removed the trailing comment. I hope his meets your requirements:
Both `Stretch` and `Fedora 28` are outdated. 
Both settings work fine on my Arch Linux setup.

PR #227, opened 2019-10-06
docs: include Arch Linux installation

juergenhoetzel:
As discussed on
https://github.com/zrepl/zrepl/pull/226#issuecomment-538737870

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -37,6 +37,9 @@ The following list may be incomplete, feel free to submit a PR with an update:
     * - MacOS
       - ``brew install zrepl``
       - Available on `homebrew <https://brew.sh>`_
+    * - Arch Linux
+      - ``yay install zrepl-bin``
+      - Available on `AUR <https://aur.archlinux.org/packages/zrepl-bin>`_
     * - Others
       -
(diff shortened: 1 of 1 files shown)


problame:
Allright!
If you're involved with Arch, would you mind making sure that their infrastructure for detecting outdated versions will detect new releases of zrepl and remind the maintainer? @johnramsden o/

johnramsden:
@problame They don't really have any infrastructure that helps people in the AUR, since they are just user-submitted. I do watch releases on [zrepl/zrepl](https://github.com/zrepl/zrepl/) though.

I'd actually prefer to be doing a from source compile, but I had difficulties getting at working last time I tried. This reminded me and I was able to get a build working. If I wanted to get your input on whether or not I am building the ideal way should I open a new issue or post the `PKGBUILD` here?

problame:
I guess a new issue would be appropriate.
The 0.2 release will switch to using Go modules, so maybe the Arch people have some abstractions around that.
For the build, the only thing left to do is provide the Go compiler flags that set zrepl's version variable, as can be easily understood from the `make release` target in the `Makefile`.

PR #236, opened 2019-10-24
zfs: actively close the copierErrChan

meilihao:


meilihao:
need to call `stdinWriter.Close()` at the same time?

problame:
The code following the two goroutines unconditionally reads from the two result channels before returning. To my understanding, this will prevent a goroutine leak if the two spawned goroutines eventually return.
Did you diagnose a leak in that code path, if so, how?

meilihao:
I read more blogs, and you are right: no leak. 

my starting point is wrong. And i am still thinking that it is a good habit to  close the channel.

my test case:
```go
func main() {
	defer func() {
		time.Sleep(time.Second)
		fmt.Println("the number of goroutines: ", runtime.NumGoroutine())
	}()

	copierErrChan := make(chan error)
	go func() {
		copierErrChan <- errors.New("test")
		close(copierErrChan)
	}()

	copierErr := <-copierErrChan
	fmt.Printf("copierErr: %T %s", copierErr, copierErr)
}
```
output:
```txt
copierErr: *errors.errorString test
the number of goroutines:  1
```

problame:
ISTM that the one goroutine in your example is the main function's goroutine.
AFAIK the garbage collector will clean up the channel if no goroutine references it, so I don't see a reason why we should add the `close` in the function referenced in your PR.

PR #233, opened 2019-10-20
rpc/dataconn + build: support GOOS={solaris,illumos}

problame:

PR #239, opened 2019-10-30
systemd + ssh-transport + go-netssh bump

### LINKED ISSUE #237: ssh+stdinserver does not work with @privileged SystemCallFilter
On setting up zrepl for the first time, I had a lot of difficulty getting replication going using ssh + stdinserver. All the logs would give me was:

```
zrepl[12230]: [redacted][rpc.ctrl]: cannot connect err="ssh exited unexpectedly with exit status 0" invocation="1"
```

Running SSH the way zrepl would call it would successfully connect and show output from the stdinserver on the remote end. I eventually ended up debugging this by calling `strace -f -p <zrepl pid>`, finding the PID of the forked ssh process (potential future reader: look for the `execve` syscall) and then filtering down to just the calls from that SSH process.

That led me to seeing the ssh process was being forcibly terminated. (I unfortunately no longer have the original strace output so I can't look up the exact lines anymore)

I tracked this down to the systemd `SystemCallFilter` configuration shipped for Ubuntu/Debian (this is on Ubuntu 18.04, for what it's worth). By default this is:

```
SystemCallFilter=~ @mount @cpu-emulation @keyring @module @obsolete @privileged @raw-io @debug @clock @resources
```

Removing `@privileged` was required to successfully replicate. I've done so with:

```
# /etc/systemd/system/zrepl.service.d/override.conf
[Service]
SystemCallFilter=
SystemCallFilter=~ @mount @cpu-emulation @keyring @module @obsolete @raw-io @debug @clock @resources
```

I would suggest removing `@privileged` from the list by default and can make a PR for that if desired.

Alternatively, this could be documented with steps to turn it off for those users wanting to use ssh+stdinserver so `@privileged` can remain by default (presumably this works for people using the plain TCP transports?) but that seems a little too much friction for new users in my opinion. Would be happy to make documentation for that as well though, if this route is desired.

problame:
#233 is based on this PR

I expect the changes here to fix #237 , copying the test plan from #237 to this PR.

---

@zoni will you have capacity to do the test plan until Wednesday next week?

Allright, go module issues resolved, commit  `ff79668` should fix your issue.
Steps to test:

- [ ] Remove the `override.conf`, i.e. use the package-provided systemd file
- [ ] `systemctl daemon-reload`
- [ ] Remove the lines for the target SSH server from your `known_hosts` file
- [ ] ( Also check for IPv4 and IPv6 addresses of that server, not just its hostname)
- [ ] Download the go1.13-built zrepl binary for linux-amd-64
- [ ] `systemctl stop zrepl`
- [ ] Copy the binary to /usr/bin/zrepl
- [ ] `chmod +x /usr/bin/zrepl`
- [ ] `systemctl restart zrepl`
- [ ] Run `zrepl version` to confirm you are running `ff79668`
- [ ] Verify that the zrepl error log contains a meaningful error message, along the lines of 'Host key verification failed`.

To download the go1.13-built zrepl binary, click on the status button next to the commit above and click on the `zrepl/publish-ci-artifacts` status: image


problame:
merged in ff rollup together with #233

PR #242, opened 2019-11-01
Add OmniOS (illumos distribution) to list of OSs

citrus-it:
A native `zrepl` package was recently added to [OmniOS](https://omnios.org) (an illumos distribution). It is currently built with Go 1.12, the Solaris arch and a few patches. In the future, it will move to Go 1.13 and the illumos arch.

Build recipe is here: https://github.com/omniosorg/omnios-extra/tree/master/build/zrepl

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -49,6 +49,9 @@ The following list may be incomplete, feel free to submit a PR with an update:
     * - Debian + Ubuntu
       - ``apt install zrepl``
       - APT repository config :ref:`see below <installation-apt-repos>`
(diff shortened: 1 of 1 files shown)


citrus-it:
Oh, and there is a sample SMF file in that source repo in case it is of use.

problame:
Great! Be prepared that #233 will refactor the Makefile a little, see 73463198bcb1a274a84d2cb6765cd3a1c3e444f6 (that commit id is likely going to get stale due to a rebase, though).

PR #259, opened 2019-12-28
{resumable,encrypted,hold-protected} send-recv, last-received-hold

problame:
From the main commit message:

new features: {resumable,encrypted,hold-protected} send-recv, last-received-hold

- **Resumable Send & Recv Support**
  No knobs required, automatically used where supported.
- **Hold-Protected Send & Recv**
  Automatic ZFS holds to ensure that we can always resume a replication step.
- **Encrypted Send & Recv Support** for OpenZFS native encryption.
  Configurable at the job level, i.e., for all filesystems a job is responsible for.
- **Receive-side hold on last received dataset**
  The counterpart to the replication cursor bookmark on the send-side.
  Ensures that incremental replication will always be possible between a sender and receiver.

Design Doc
----------

`replication/design.md` doc describes how we use ZFS holds and bookmarks to ensure that a single replication step is always resumable.

The replication algorithm described in the design doc introduces the notion of job IDs (please read the details on this design doc).
We reuse the job names for job IDs and use `JobID` type to ensure that a job name can be embedded into hold tags, bookmark names, etc.
This might BREAK CONFIG on upgrade.

Protocol Version Bump
---------------------

This commit makes backwards-incompatible changes to the replication/pdu protobufs.
Thus, bump the version number used in the protocol handshake.

Replication Cursor Format Change
--------------------------------

The new replication cursor bookmark format is: `#zrepl_CURSOR_G_${this.GUID}_J_${jobid}`
Including the GUID enables transaction-safe moving-forward of the cursor.
Including the job id enables that multiple sending jobs can send the same filesystem without interfering.
The `zrepl migrate replication-cursor:v1-v2` subcommand can be used to safely destroy old-format cursors once zrepl has created new-format cursors.

Changes in This Commit
----------------------

- package zfs
  - infrastructure for holds
  - infrastructure for resume token decoding
  - implement a variant of OpenZFS's `entity_namecheck` and use it for validation in new code
  - ZFSSendArgs to specify a ZFS send operation
    - validation code protects against malicious resume tokens by checking that the token encodes the same send parameters that the send-side would use if no resume token were available (i.e. same filesystem, `fromguid`, `toguid`)
  - RecvOptions support for `recv -s` flag
  - convert a bunch of ZFS operations to be idempotent
    - achieved through more differentiated error message scraping / additional pre-/post-checks

- package replication/pdu
  - add field for encryption to send request messages
  - add fields for resume handling to send & recv request messages
[...]

PR #265, opened 2020-01-18
serve tcp: CIDR / subnet based ACLs + auto client-identities

Smithx10:
I've manually tested this,  and believe this achieves the desired behavior.  

Client Config:
```
[root@backup01 ~]# cat zrepl.yaml
jobs:
- name: PGaaSBackup
  type: sink
  serve:
    type: tcp
    listen: :8888
    clients:
     "*": "*"
```

End Result: 
```
54d-1eb8-451e-abce-ec8224788442/data/zrepl
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.28.61                                                           12.9M   999G   219K  none
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.28.61/zones                                                     12.7M   999G   219K  none
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.28.61/zones/65f134cf-ae7d-cf46-9664-edbed718fb0e                12.5M   999G   219K  none
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.28.61/zones/65f134cf-ae7d-cf46-9664-edbed718fb0e/data           12.3M   999G   219K  none
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.28.61/zones/65f134cf-ae7d-cf46-9664-edbed718fb0e/data/manatee   12.1M   999G  12.1M  none
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.29.247                                                          13.2M   999G   219K  none
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.29.247/zones                                                    13.0M   999G   219K  none
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.29.247/zones/3ec51b26-bcce-e6b1-a699-eeb5ec11010c               12.8M   999G   219K  none
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.29.247/zones/3ec51b26-bcce-e6b1-a699-eeb5ec11010c/data          12.5M   999G   219K  none
zones/a14bd54d-1eb8-451e-abce-ec8224788442/data/zrepl/50.50.29.247/zones/3ec51b26-bcce-e6b1-a699-eeb5ec11010c/data/manatee  12.3M   999G  12.1M  none

```

### DIFF
--- a/transport/tcp/serve_tcp.go
+++ b/transport/tcp/serve_tcp.go
@@ -11,39 +11,6 @@ import (
 	"github.com/zrepl/zrepl/util/tcpsock"
 )
 
-type ipMapEntry struct {
-	ip    net.IP
-	ident string
-}
-
-type ipMap struct {
-	entries []ipMapEntry
-}
-
-func ipMapFromConfig(clients map[string]string) (*ipMap, error) {
-	entries := make([]ipMapEntry, 0, len(clients))
-	for clientIPString, clientIdent := range clients {
-		clientIP := net.ParseIP(clientIPString)
-		if clientIP == nil {
-			return nil, errors.Errorf("cannot parse client IP %q", clientIPString)
-		}
-		if err := transport.ValidateClientIdentity(clientIdent); err != nil {
-			return nil, errors.Wrapf(err, "invalid client identity for IP %q", clientIPString)
-		}
-		entries = append(entries, ipMapEntry{clientIP, clientIdent})
-	}
-	return &ipMap{entries: entries}, nil
-}
-
-func (m *ipMap) Get(ip net.IP) (string, error) {
-	for _, e := range m.entries {
-		if e.ip.Equal(ip) {
-			return e.ident, nil
-		}
-	}
-	return "", errors.Errorf("no identity mapping for client IP %s", ip)
-}
-
 func TCPListenerFactoryFromConfig(c *config.Global, in *config.TCPServe) (transport.AuthenticatedListenerFactory, error) {
 	clientMap, err := ipMapFromConfig(in.Clients)
 	if err != nil {
@@ -69,10 +36,13 @@ func (f *TCPAuthListener) Accept(ctx context.Context) (*transport.AuthConn, erro
 	if err != nil {
 		return nil, err
 	}
-	clientIP := nc.RemoteAddr().(*net.TCPAddr).IP
-	clientIdent, err := f.clientMap.Get(clientIP)
+	clientAddr := &net.IPAddr{
+		IP:   nc.RemoteAddr().(*net.TCPAddr).IP,
+		Zone: nc.RemoteAddr().(*net.TCPAddr).Zone,
+	}
+	clientIdent, err := f.clientMap.Get(clientAddr)
 	if err != nil {
-		transport.GetLogger(ctx).WithField("ip", clientIP).Error("client IP not in client map")
+		transport.GetLogger(ctx).WithField("ipaddr", clientAddr).Error("client IP not in client map")
 		nc.Close()
 		return nil, err
 	}
--- a/transport/tcp/serve_tcp_ipmap.go
+++ b/transport/tcp/serve_tcp_ipmap.go
@@ -1,0 +1,166 @@
+package tcp
+
+import (
+	"bytes"
+	"fmt"
+	"net"
+	"sort"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/zrepl/zrepl/transport"
+	"golang.org/x/sys/unix"
+)
+
+type ipMapEntry struct {
+	subnet *net.IPNet
+	// ident is always not empty
+	ident string
+	// zone may be empty (e.g. for IPv4)
+	zone string
+}
+
+type ipMap struct {
+	entries []*ipMapEntry
+}
+
+func ipMapFromConfig(clients map[string]string) (*ipMap, error) {
+
+	entries := make([]*ipMapEntry, 0, len(clients))
+	for clientInput, clientIdent := range clients {
+		userIPMapEntry, err := newIPMapEntry(clientInput, clientIdent)
+		if err != nil {
+			return nil, errors.Wrapf(err, "cannot not parse %q:%q", clientInput, clientIdent)
+		}
+
+		entries = append(entries, userIPMapEntry)
+	}
+
+	sort.Sort(byPrefixlen(entries))
+
+	return &ipMap{entries: entries}, nil
+}
+
+func (m *ipMap) Get(ipAddr *net.IPAddr) (string, error) {
+	for _, e := range m.entries {
+		if e.zone != ipAddr.Zone {
+			continue
+		}
+		if e.subnet.Contains(ipAddr.IP) {
+			return zfsDatasetPathComponentCompatibleRepresentation(e.ident, ipAddr), nil
+		}
+	}
+	return "", errors.Errorf("no identity mapping for client IP: %s%%%s", ipAddr.IP, ipAddr.Zone)
+}
+
+var ipv6FullySpecifiedMask = bytes.Repeat([]byte{0xff}, net.IPv6len)
+
+func newIPMapEntry(input string, ident string) (*ipMapEntry, error) {
+	ip := input
+	var zone string
+
+	ipZoneSplit := strings.SplitN(input, "%", 2)
+	if len(ipZoneSplit) > 1 {
+		ip = ipZoneSplit[0]
+		zone = ipZoneSplit[1]
+	}
+
+	_, subnet, err := net.ParseCIDR(ip)
+	if err != nil {
+		// expect full IP, no '*' placeholder expansion
+
+		if strings.Count(ident, "*") != 0 {
+			return nil, fmt.Errorf("non-CIDR matches must not contain '*' placeholder")
+		}
+
+		if err := transport.ValidateClientIdentity(ident); err != nil {
+			return nil, errors.Wrapf(err, "invalid client identity %q for IP %q", ident, ip)
+		}
+
+		parsedIP := net.ParseIP(ip)
+		if parsedIP == nil {
+			return nil, errors.Wrapf(err, "invalid client address %q", ip)
+		}
+		parsedIP = parsedIP.To16()
+
+		return &ipMapEntry{
+			subnet: &net.IPNet{
+				IP:   parsedIP,
+				Mask: ipv6FullySpecifiedMask,
+			},
+			zone:  zone,
+			ident: ident,
+		}, nil
+	}
+
+	// expect CIDR and '*' placeholder expansion
+
+	if strings.Count(ident, "*") != 1 {
+		err = fmt.Errorf("CIDRs require 1 IP placeholder")
+		return nil, errors.Wrapf(err, "invalid client identity %q for IP %q", ident, ip)
+	}
+
+	longestIPAddr := net.IPAddr{
+		IP:   net.IP(bytes.Repeat([]byte{0xff}, net.IPv6len)),
+		Zone: strings.Repeat("i", unix.IFNAMSIZ),
+	}
+	longestIdent := zfsDatasetPathComponentCompatibleRepresentation(ident, &longestIPAddr)
+
+	if err := transport.ValidateClientIdentity(longestIdent); err != nil {
+		return nil, errors.Wrapf(err, "invalid client identity for IP %q", ip)
+	}
+
+	ones, _ := subnet.Mask.Size()
+	preExpansionAddrlen := len(subnet.IP) * 8
+	expanded := subnet.IP.To16()
+	postExpansionAddrlen := len(expanded) * 8
+	return &ipMapEntry{
+		subnet: &net.IPNet{
+			IP:   expanded,
+			Mask: net.CIDRMask(postExpansionAddrlen-preExpansionAddrlen+ones, postExpansionAddrlen),
+		},
+		zone:  zone,
+		ident: ident,
+	}, nil
+}
+
+func zfsDatasetPathComponentCompatibleRepresentation(identityWithWildcard string, addr *net.IPAddr) string {
+	// If a Zone exists we append it after the IP using a "-" because "%"
+	// is a zfs dataset forbidden char.
+	if addr.Zone != "" {
+		return strings.Replace(identityWithWildcard, "*", addr.IP.String()+"-"+addr.Zone, 1)
+	}
+	// newIPMapEntry validates that the line contains exactly one "*"
+	return strings.Replace(identityWithWildcard, "*", addr.IP.String(), 1)
+}
+
+func (e *ipMapEntry) String() string {
+	return fmt.Sprintf("&ipMapEntry{subnet=%q, ident=%q, zone=%q}", e.subnet.String(), e.ident, e.zone)
+}
+
+func (e *ipMapEntry) PrefixLen() int {
+	ones, bits := e.subnet.Mask.Size()
+	if bits != net.IPv6len*8 {
+		panic(fmt.Sprintf("impl error: we represent all addresses as 16byte internally ones=%v bits=%v: %s", ones, bits, e))
+	}
+	return ones
+}
+
+// newtype to support sorting by prefixlength
+type byPrefixlen []*ipMapEntry
+
+func (m byPrefixlen) Len() int { return len(m) }
+func (m byPrefixlen) Less(i, j int) bool {
+
+	if m[i].PrefixLen() != m[j].PrefixLen() {
+		return m[i].PrefixLen() > m[j].PrefixLen()
+	}
+
+	addrCmp := bytes.Compare(m[i].subnet.IP.To16(), m[j].subnet.IP.To16())
+	if addrCmp != 0 {
+		return addrCmp < 0
+	}
+
+	return strings.Compare(m[i].zone, m[j].zone) < 0
+}
+func (m byPrefixlen) Swap(i, j int) { m[i], m[j] = m[j], m[i] }
--- a/transport/tcp/serve_tcp_ipmap_test.go
+++ b/transport/tcp/serve_tcp_ipmap_test.go
@@ -1,0 +1,204 @@
+package tcp
+
+import (
+	"net"
+	"os"
+	"testing"
+
+	"github.com/kr/pretty"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIPMap(t *testing.T) {
+	type testCaseExpect struct {
+		expectNoMapping bool
+		expectIdent     string
+	}
+	type testCase struct {
+		name          string
+		config        map[string]string
+		expectInitErr bool
+		expect        map[string]testCaseExpect
+	}
+
+	cases := []testCase{
+		{
+			name:          "regular ips",
+			expectInitErr: false,
+			config: map[string]string{
+				"192.168.123.234": "ident1",
+				"192.168.20.23":   "ident2",
+			},
+			expect: map[string]testCaseExpect{
+				"192.168.123.234": {expectIdent: "ident1"},
+				"192.168.20.23":   {expectIdent: "ident2"},
+				"192.168.20.10":   {expectNoMapping: true},
+			},
+		},
+		{
+			name:          "full wildcard",
+			expectInitErr: false,
+			config: map[string]string{
+				"0.0.0.0/0": "*",
+				"0::/0":     "*",
+			},
+			expect: map[string]testCaseExpect{
+				"10.123.234.24": {expectIdent: "10.123.234.24"},
+				// '[' and ']' are forbiddenin dataset names
+				"fe80::23:42": {expectIdent: "fe80::23:42"},
+			},
+		},
+		{
+			name:          "longest prefix matching",
+			expectInitErr: false,
+			config: map[string]string{
+				"10.1.2.3":              "specific-host",
+				"10.1.2.0/24":           "subnet-one-two-*",
+				"10.1.1.0/24":           "subnet-one-one-*",
+				"10.1.0.0/24":           "subnet-one-zero-*",
+				"10.1.0.0/16":           "subnet-one-*",
+				"fde4:8dba:82e1:1::1":   "v6-48-1-specialhost",
+				"fde4:8dba:82e1::/48":   "v6-48-*",
+				"fde4:8dba:82e1:1::/64": "v6-64-1-*",
+				"fde4:8dba:82e1:2::/64": "v6-64-2-*",
+			},
+			expect: map[string]testCaseExpect{
+				"10.1.2.3":             {expectIdent: "specific-host"},
+				"10.1.2.1":             {expectIdent: "subnet-one-two-10.1.2.1"},
+				"10.1.1.1":             {expectIdent: "subnet-one-one-10.1.1.1"},
+				"10.1.0.23":            {expectIdent: "subnet-one-zero-10.1.0.23"},
+				"10.1.3.1":             {expectIdent: "subnet-one-10.1.3.1"},
+				"10.2.1.1":             {expectNoMapping: true},
+				"fde4:8dba:82e1:1::1":  {expectIdent: "v6-48-1-specialhost"},
+				"fde4:8dba:82e1:23::1": {expectIdent: "v6-48-fde4:8dba:82e1:23::1"},
+				"fde4:8dba:82e1:1::2":  {expectIdent: "v6-64-1-fde4:8dba:82e1:1::2"},
+				"fde4:8dba:82e1:2::1":  {expectIdent: "v6-64-2-fde4:8dba:82e1:2::1"},
+				"fde4:8dba:82e2::1":    {expectNoMapping: true},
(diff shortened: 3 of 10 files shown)


Smithx10:
I believe I could take this a bit further, and ACL on subnets...  but if you are using TCP I'd imagine the zrepl users network is isolated or they've handled the encapsulation in some other manner.

Smithx10:
One way to identify the Snapshots is to include a human friendly named in prefix.  This makes finding the snapshots easier and is defined in the pusher.

problame:
Thanks for starting to work on this, but if I merge this feature, I'd like to see subnet filtering.
I pushed a test suite for ipMap to this branch which defines what I'd like to see - featurewise.

It doesn't pass yet, and if you have the time, I'd appreciate if you do the impl work.
Please use https://github.com/yl2chen/cidranger for longest prefix matching.

What do you think?

Smithx10:
@problame  Thanks for the quick feed back.  Sounds like a plan.  Hopefully while I'm watching the UFC tonight I can finish this up.  That library looks cool!

Smithx10:
Errr,  I thought the test with a  "IPv6 link-local address with interface name" was a typo.  After some googling,  I can't find an example on how to parse that properly.    If it's in the stdlib I can't find it,  did some googling too :(

transport/tcp/serve_tcp.go
```diff
@@ -69,10 +36,13 @@ func (f *TCPAuthListener) Accept(ctx context.Context) (*transport.AuthConn, erro
...
-	clientIP := nc.RemoteAddr().(*net.TCPAddr).IP
-	clientIdent, err := f.clientMap.Get(clientIP)
+	clientAddr := &net.IPAddr{
+		IP:   nc.RemoteAddr().(*net.TCPAddr).IP,
+		Zone: nc.RemoteAddr().(*net.TCPAddr).Zone,
+	}
+	clientIdent, err := f.clientMap.Get(clientAddr)
 	if err != nil {
-		transport.GetLogger(ctx).WithField("ip", clientIP).Error("client IP not in client map")
+		transport.GetLogger(ctx).WithField("ip", clientAddr.IP).Error("client IP not in client map")
```
problame: please log the full net.IPAddr here

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -0,0 +1,113 @@
...
+					return ipAddr.IP.String(), nil
+				}
+				return e.ident + ipAddr.IP.String(), nil
+			}
+		}
+	}
+	return "", errors.Errorf("no identity mapping for client IP: %s%%%s", ipAddr.IP, ipAddr.Zone)
+}
+
+func handleIPV6(input string) (net.IP, *net.IPNet, string) {
```
problame: a more expressive name would be `parseUserIPMatchExpr` or similar.
Also, I would rather see this return a pointer to a struct, and have the pointer be nil if parsing fails

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -0,0 +1,113 @@
...
+			contains := e.subnet.Contains(ipAddr.IP)
+			if contains {
+				if e.ident == "willbeclientip" {
+					return ipAddr.IP.String(), nil
+				}
+				return e.ident + ipAddr.IP.String(), nil
+			}
+		}
+	}
+	return "", errors.Errorf("no identity mapping for client IP: %s%%%s", ipAddr.IP, ipAddr.Zone)
```
problame: again, `ipAddr.String()` takes care of zone formatting if `.Zone != ""`

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func ipMapFromConfig(clients map[string]string) (*ipMap, error) {
+	entries := make([]ipMapEntry, 0, len(clients))
+	for clientInput, clientIdent := range clients {
+		clientIP, clientCIDR, zone := handleIPV6(clientInput)
+
+		if clientCIDR != nil {
+			if len(clientIdent) == 1 && clientIdent == "*" {
+				clientIdent = "willbeclientip"
```
problame: `"willbeclientip"` would be a legitimate user-specified client identity
=> you cannot use that as a sentinel value

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+			if len(clientIdent) > 1 && clientIdent[len(clientIdent)-1:] == "*" {
+				clientIdent = strings.TrimRight(clientIdent, "*")
```
problame: Please add a test case that tests a mapping a la `"10.1.2.0/24":"my-*-server"`, which should evaluate like this: `"10.1.2.3":"my-10.1.2.3-server"`.

Also, there should  be a test cases for:

```
"10.1.2.3":"too-lazy-to-repeat-left-hand-side-*" => 10.1.2.3 maps to "too-lazy-to-repeat-left-hand-side-10.1.2.3*

"10.1.2.0/24":"expression-without-star" => expectInitErr = true
```

transport/tcp/serve_tcp_ipmap_test.go
```diff
@@ -0,0 +1,125 @@
...
+			},
+		},
+		{
+			name:          "longest prefix matching",
+			expectInitErr: false,
+			config: map[string]string{
+				"10.1.2.3":    "specific-host",
+				"10.1.2.0/24": "subnet-one-two-*",
+				"10.1.1.0/24": "subnet-one-one-*",
+				"10.1.0.0/16": "subnet-one-*",
```
problame: should also have:
```
"10.1.0.0/24":`subnet-one-zero-*"
=> map 10.1.0.23 to "subnet-one-zero-10.1.0.23"
```

transport/tcp/serve_tcp_ipmap_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIPMap(t *testing.T) {
+	type testCaseExpect struct {
+		expectErr   bool
```
problame: should rename this field to expectNoMapping

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+type ipMap struct {
+	entries []ipMapEntry
+}
+
+func ipMapFromConfig(clients map[string]string) (*ipMap, error) {
+	entries := make([]ipMapEntry, 0, len(clients))
+	for clientInput, clientIdent := range clients {
+		clientIP, clientCIDR, zone := handleIPV6(clientInput)
```
problame: error handling for invalid client input missing

transport/tcp/serve_tcp_ipmap_test.go
```diff
@@ -0,0 +1,125 @@
...
+				// v6 matching
+				"fe80::23:42%eth1": {expectIdent: "san-fe80::23:42"},
+				"fe80::23:42%eth2": {expectErr: true},
+
+				"2001:0db8:85a3:0000:0000:8a2e:0370:7334":      {expectIdent: "aspecifichost"},
+				"2001:0db8:85a3:0000:0000:8a2e:0370:7334%eth1": {expectIdent: "aspecifichost"},
+				"2001:0db8:85a3:0000:0000:8a2e:0370:7334%eth2": {expectErr: true},
+			},
+		},
+	}
```
problame: we need test cases for invalid user input, i.e. those where `expectInitErr: true`.

separate ones for each kind of invalid input!

Smithx10:
Ahhh Sorry about the force push,  I reset and forgot :(

transport/tcp/serve_tcp.go
```diff
@@ -42,7 +42,7 @@ func (f *TCPAuthListener) Accept(ctx context.Context) (*transport.AuthConn, erro
 	}
 	clientIdent, err := f.clientMap.Get(clientAddr)
 	if err != nil {
-		transport.GetLogger(ctx).WithField("ip", clientAddr.IP).Error("client IP not in client map")
+		transport.GetLogger(ctx).WithField("ipaddr", clientAddr).Error("client ipaddr not in client map")
```
problame: "client IP" is just fine here

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -17,33 +18,18 @@ type ipMapEntry struct {
...
-		if clientCIDR != nil {
-			if len(clientIdent) == 1 && clientIdent == "*" {
-				clientIdent = "willbeclientip"
-			}
-			if len(clientIdent) > 1 && clientIdent[len(clientIdent)-1:] == "*" {
-				clientIdent = strings.TrimRight(clientIdent, "*")
-			}
+		userIPMapEntry, err := newiPMapEntry(clientInput, clientIdent)
+		if err != nil {
+			return nil, err
```
problame: add context to the returned error by wrapping it:
```
errors.Wrapf(err, "parse %q:%q", clientInput, clientIdent)
```

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -60,38 +46,62 @@ func (m *ipMap) Get(ipAddr *net.IPAddr) (string, error) {
 			return e.ident, nil
 		}
 		if e.subnet != nil {
-			contains := e.subnet.Contains(ipAddr.IP)
-			if contains {
-				if e.ident == "willbeclientip" {
-					return ipAddr.IP.String(), nil
-				}
-				return e.ident + ipAddr.IP.String(), nil
+			if e.subnet.Contains(ipAddr.IP) {
```
problame: take advantage of short-circuiting by making the encloding if-condition `e.subnet != nil && e.subnet.Contains(ipAddr.IP)` => makes the code more concise

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -60,38 +46,62 @@ func (m *ipMap) Get(ipAddr *net.IPAddr) (string, error) {
...
+				return strings.Replace(e.ident, "*", ipAddr.IP.String(), 1), nil
```
problame: maybe put an assertion or at least a comment here that it's guaranteed that e.ident will contain at most one "*"

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -60,38 +46,62 @@ func (m *ipMap) Get(ipAddr *net.IPAddr) (string, error) {
...
+		err = fmt.Errorf("CIDRs require 1 IP placeholder")
+		return nil, errors.Wrapf(err, "invalid client identity for IP %q", ip)
+	}
+	if len(ident) == 1 {
+		validationString = strings.Replace(ident, "*", "pass", 1)
+	} else {
+		validationString = strings.Replace(ident, "*", "", 1)
+	}
+
+	if err := transport.ValidateClientIdentity(validationString); err != nil {
```
problame: I think it would be best to have the validation string be the maximum-length of an address's `.String()`. perhaps by setting all bits in the address to 1?

About the `zone`: the `%` character is not allowed in a dataset path component (see zfs function `int
zfs_component_namecheck(const char *path, namecheck_err_t *why, char *what)`).

The upcoming big-feature branch #259  brings in better dataset path validation, so we'll start using that in `transport.ValidateClientIdentity`.
=> we will need a way to deal with zones (perhaps detect and forbid them? or only allow them in full mappings?)
==> I'll leave that up to you, just make sure to turn your decision into a test case (for both positive and negatve variants).

transport/tcp/serve_tcp_ipmap_test.go
```diff
@@ -71,26 +73,63 @@ func TestIPMap(t *testing.T) {
...
+		},
+		{
+			name:          "invalid user input: v4 ip with identifier containing *",
+			expectInitErr: true,
+			config: map[string]string{
+				"192.168.1.2": "db-*",
+			},
+		},
+		{
+			name:          "invalid user input: v4 subnet without identifier containing *",
```
problame: s/identifier/identity, also in test names above and below

transport/tcp/serve_tcp_ipmap_test.go
problame: great work!

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -60,38 +46,62 @@ func (m *ipMap) Get(ipAddr *net.IPAddr) (string, error) {
...
 }
 
-func handleIPV6(input string) (net.IP, *net.IPNet, string) {
+func newiPMapEntry(input string, ident string) (*ipMapEntry, error) {
 	ip := input
 	var zone string
+	validationString := ident
 	if strings.Contains(input, "%") {
 		ip = strings.Split(input, "%")[0]
 		zone = strings.Split(input, "%")[1]
```
problame: do this with `SplitN(input, "%", 1)`

transport/tcp/serve_tcp_ipmap_test.go
```diff
@@ -106,13 +145,13 @@ func TestIPMap(t *testing.T) {
 				require.NotNil(t, m)
 			}
 			for input, expect := range c.expect {
-				ip, _, zone := handleIPV6(input)
-				require.NotNil(t, ip)
+				ipMapEntry, _ := newiPMapEntry(input, "test")
```
problame: comment: reuse newiPMapEntry to parse test case input

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -60,38 +46,62 @@ func (m *ipMap) Get(ipAddr *net.IPAddr) (string, error) {
...
+			if e.subnet.Contains(ipAddr.IP) {
+				return strings.Replace(e.ident, "*", ipAddr.IP.String(), 1), nil
 			}
 		}
 	}
 	return "", errors.Errorf("no identity mapping for client IP: %s%%%s", ipAddr.IP, ipAddr.Zone)
 }
 
-func handleIPV6(input string) (net.IP, *net.IPNet, string) {
+func newiPMapEntry(input string, ident string) (*ipMapEntry, error) {
```
problame: s/newiPMapEntry/newIPMapEntry/

transport/tcp/serve_tcp_ipmap.go
Smithx10: The current behavior allows for people to use a Zone, and it will be ACL'd against but their is no representation of this in the dataset path.  I imagine we could just do what DNS does here and in the path convert to "-".  It's not a forbidden char. ```const FORBIDDEN = "@#|\t<>*"```  I'll wire something up and see if we like it or not.   This would be handy if someone was doing the same ipv6 subnet ranges but with different zones.  They could prefix "web-eth0-*  but we could just handle that for them.   Probably a better user experience, and is more consistent.

Smithx10: "I think it would be best to have the validation string be the maximum-length of an address's .String(). perhaps by setting all bits in the address to 1?"

I'm not following ( I'm kind of a noob. ),  isn't the maximum length of the validationString dependent on the identity provided in the config?  We already know that IP addresses coming in are valid. We must verify that "subnet-one-two-" of  user input "subnet-one-two-*"  validates cleanly.   
```
				"10.1.2.3":    "specific-host",
				"10.1.2.0/24": "subnet-one-two-*",
				"10.1.1.0/24": "subnet-one-one-*",
				"10.1.0.0/24": "subnet-one-zero-*",
```

Speaking of that,  I'll add some tests to assert that all of the forbidden chars are actually forbidden.

problame: What I mean is that _for upfron validation when parsing the config input_ we should use the `identity-with-*` from the config, replace `*` with the longest-possible expansion of the left-hand side (`255.255.255.255` for IPv4, `2001:0db8:0000:0000:0000:8a2e:0370:7334` or all `ffff`s for IPv6, plus a maximum zone length if there is any), and feed the result of that expansion to the validation function.

problame: > Speaking of that, I'll add some tests to assert that all of the forbidden chars are actually forbidden.

As stated: the actual ZFS implementation (unlike what we do in zrepl's current master branch, see #259 for changes to that) is a whitelist rather than a blacklist => the ZFS function `int zfs_component_namecheck(const char *path, namecheck_err_t *why, char *what)` is authoritative, the implementation in zrepl's current master is incomplete. #259 reimplements `zfs_component_namecheck` in Go, and we'll switch to #259 's impl once that is merged)

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -45,22 +46,28 @@ func (m *ipMap) Get(ipAddr *net.IPAddr) (string, error) {
...
-				return strings.Replace(e.ident, "*", ipAddr.IP.String(), 1), nil
+		if e.subnet != nil && e.subnet.Contains(ipAddr.IP) {
+			// If a Zone exists we append it after the IP using a "-" because "%"
+			// is a zfs dataset forbidden char.
+			if e.zone != "" {
+				return strings.Replace(e.ident, "*", ipAddr.IP.String()+"-"+e.zone, 1), nil
 			}
+			// Because of actions taken earlier Line #84 Subnets will
+			// only contain 1 "*" for replacement.
+			return strings.Replace(e.ident, "*", ipAddr.IP.String(), 1), nil
```
problame: I think we should incorporate the logic in the if branch into a separate helper function, e.g. `zfsDatasetPathComponentCompatibleRepresentation(addr *net.IpAddr)`

transport/tcp/serve_tcp_ipmap_test.go
```diff
@@ -105,47 +106,99 @@ func TestIPMap(t *testing.T) {
...
+			},
+		},
 	}
 
 	for i := range cases {
 		c := cases[i]
 		t.Run(c.name, func(t *testing.T) {
 			m, err := ipMapFromConfig(c.config)
 			if c.expectInitErr {
+				fmt.Println(err)
```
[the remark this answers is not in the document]
Smithx10: eeek,  I was doing that while testing the tests. Doh.

transport/tcp/serve_tcp_ipmap.go
Smithx10: Ahhh,  that makes more sense now.  Thanks for explaining that.  I didn't take into consideration that a zfs dataset has a max size of 255.    Should we also validate that "root_fs"+"ident"+"ip"+"-zone" <= 255?  I don't think (didn't see) where zfs.NewDatasetPath checks this.

Smithx10: I added the following test,  and it passed. (258 chars)
				"10.1.4.0/24": "zwE0tyDzO5mouG3yfB8yCepc6day3ShQzWR2ZSG6J6d2dcBYaWKepCzp9aLtrNg8KoamJM0liqFtZdJJx4L9WUabqEtFy2G45EClbNHhOYxBKCV3shkDi6xtGJVRUOvPhINjLh95uMyHdFvDjBhYkhnobzk64fRgtcy8A4Ad2EorZdKUJ73IrAMNcZFUjuURgHmw9ZBSjlLjIb7vwrjFjI8DWT8xwNu6c3Jcdu7ELihb7t7y4gF254nqDeMESLa1-*",

problame: > Should we also validate that "root_fs"+"ident"+"ip"+"-zone" <= 255? I don't think (didn't see) where zfs.NewDatasetPath checks this.

That'll be the job of `transport.ValidateClientIdentity`.
#259 will bring `ComponentNamecheck`, which is what it should use for that purpose.
I assume I'll merge this PR after #259 , so if we rebase this PR after #259 is merged, we can fix that then. I'll add a TODO.

problame: (Keep above test in the test suite anyways!)

problame:
- [ ] update `transport.ValidateClientIdentity` to use #259 `zfs.ComponentNamecheck` (wait until #259 is merged)

transport/tcp/serve_tcp_ipmap.go
Smithx10: So besides this last validation,  everything else look ok?    How would you look like to move forward,  should we wait until 259 is merged,  and then handle that last little bit?   It look's like TCPListenerFactoryFromConfig would have needed the root_fs passed in also,  unless I'm missing something.    Is there a simpler way to get that piece of config from the job?    Thanks again for review and help.

problame:
> So besides this last validation, everything else look ok? How would you look like to move forward, should we wait until 259 is merged, and then handle that last little bit? It look's like TCPListenerFactoryFromConfig would have needed the root_fs passed in also, unless I'm missing something. Is there a simpler way to get that piece of config from the job? Thanks again for review and help.

Pretty much. Please go through each of the open review items ("conversation") and mark them as resolved if you feel like they are (in particular those marked as Outdated; GitHub's autodetection isn't always the best).
I'll wait until #259 is merged, then ask you to rebase this branch onto master, and give it one last round of review before merging it.

If you'd like to speed the #259 merge up, you could also give that branch a test run on your own 
deployment, and maybe have this feature cherry-picked on top of it (in your private copy).

- [ ] Also, this feature needs to be documented in `docs/transports.rst` for the `tcp` transport:
  - explanation of the CIDR matching
    - explanation of `*` expansion (and that it's required if CIDR matching is used)
  - integrate into example configuration (based on the existing one)

problame:
@Smithx10 are you still going to make the requested documentation changes (see TODO item in previous message)? They should be fairly orthogonal to the `transport.ValidateClientIdentity` changes which I'll do after the #259 is merged.

Smithx10:
@problame Sure,  when I get some cycles I'll see about getting that done.

problame:
#259 is merged as of a few minutes ago  - I will rebase this branch, do the docs updates and merge the feature.

problame:
@Smithx10 I force-pushed my rebase + additional fixes to your branch. Would you please give it a last review pass / run this branch on your systems?
Are you able to provide feedback until this Wednesday 17:00 UTC? I want to issue the release that day.

Smithx10:
@problame  Thanks for picking up my slack,  I got pulled into something for work and just forgot all about finishing up that PR.   Much appreciated!     I'll give it a once over tonight.

transport/tcp/serve_tcp_ipmap.go
```diff
@@ -80,10 +88,15 @@ func newIPMapEntry(input string, ident string) (*ipMapEntry, error) {
...
 
+	// expect CIDR and '*' placeholder expansion
+
 	if strings.Count(ident, "*") != 1 {
 		err = fmt.Errorf("CIDRs require 1 IP placeholder")
-		return nil, errors.Wrapf(err, "invalid client identity for IP %q", ip)
+		return nil, errors.Wrapf(err, "invalid client identity %q for IP %q", ident, ip)
 	}
+
+	// FIXME use longest possible validation string
```
Smithx10: @problame 

What stops us from getting the root_fs and validating the length of the entire dataset name against  zfs.MaxDatasetNameLen?   Do we ever validate MaxLength of the entire ZFS dataset (root_fs + "/" + zfsComponent) anywhere else?

Smithx10:
Everything else look's good,  just curious about the max length.

transport/tcp/serve_tcp_ipmap.go
problame: >  Do we ever validate MaxLength of the entire ZFS dataset (root_fs + "/" + zfsComponent) anywhere else?

The sending-side filesystem name is unknown at the time we check the receiving-side config, thus anything we do will only be a best-guess.

This FIXME aims at using a maximum-length IP address string.

Smithx10: @problame  Ahhhh that makes sense.    I guess any errors for that will show up in the log and zrepl status command?

problame:
please have a look at the changes I just pushed

transport/tcp/serve_tcp_ipmap.go
problame: yes, exactly, it's a runtime error

Smithx10:
lgtm

problame:
Sry for letting this sit for so long.
I just rebased the branch onto current master and pushed it to this PR.

I have two main complaints about the implementation of `ipMap`:

- `ipMapEntry` contains both an `ip` and a `subnet`, and we have the convention that only either one is valid
 -  we put both IPv4 and IPv6 `ipMapEntry`s in the same list, sort it by prefix length, the find the most specific matching prefix

I think this can be simplified by

- having 2 lists, one for v4, one for v6
- always using a net.IPNet
  - ... and converting single addresses to net.IPNet with full prefix length in `newIPMapEntry`

What do you think?
Have you used this code in your setup?

problame:
I pushed a patch that does the unified `net.IPNet` (no more `.ip != nil XOR .subnet != nil`).
The split into two separate lists seems unnecessary given that we expand v4s to embedded-in-v6 addrs.

PR #268, opened 2020-01-30
daemon/job/build_jobs: fix validateReceivingSidesDoNotOverlap

MAFLO321:
Before this was false positiv interpreted as overlapping filesystem.
```
zpool/foo
zpool/foobar
```
This PR fixes this and marks it as non-overlapping.

- [x] Fix receiving sides not overlapping check
- [x] Add test for this


problame:
Thanks for your contribution!

PR #269, opened 2020-02-01
Fix build on FreeBSD on aarch64

woodsb02:
by updating golang:sys dependency


woodsb02:
This was original reported here:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=242456

And originally fixed upstream here:
https://go.googlesource.com/sys/+/33540a1f6037

problame:
- [x] updated the corresponding go.sum file

PR #272, opened 2020-02-09
build: add linux/arm (32bit) target

problame:
fixes #271

PR #273, opened 2020-02-09
build: add linux/armv7 target

### LINKED ISSUE #271: Support for `armhf` architecture (e.g. ARMv7 32-bit)
Are there any plans to support (i.e. provide binaries for) the `armhf` architecture (ARMv7 32-bit) on Debian/Ubuntu?

I just added the [Debian/Ubuntu APT repositories](https://zrepl.github.io/installation.html#debian-ubuntu-apt-repositories) and realised that it does not have packages for the `armhf` architecture.

I was looking into using `zrepl` on my [Helios4 NAS](https://wiki.kobol.io/helios4/intro/), which is based on a Marvell Armada 388 SoC with a ARM Cortex-A9 CPU (ARMv7 32-bit). ZFS (0.7.12) is working quite nicely on that system, and I would expect that there isn't much of a reason why `zrepl` wouldn't work their either.

MAFLO321:
fixes #271
supersedes PR #272

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ endif
 GO := go
 GOOS ?= $(shell bash -c 'source <($(GO) env) && echo "$$GOOS"')
 GOARCH ?= $(shell bash -c 'source <($(GO) env) && echo "$$GOARCH"')
+GOARM ?= $(shell bash -c 'source <($(GO) env) && echo "$$GOARM"')
 GOHOSTOS ?= $(shell bash -c 'source <($(GO) env) && echo "$$GOHOSTOS"')
 GOHOSTARCH ?= $(shell bash -c 'source <($(GO) env) && echo "$$GOHOSTARCH"')
 GO_ENV_VARS := GO111MODULE=on
@@ -25,11 +26,17 @@ GO_MOD_READONLY := -mod=readonly
 GO_BUILDFLAGS := $(GO_MOD_READONLY)
 GO_BUILD := $(GO_ENV_VARS) $(GO) build $(GO_BUILDFLAGS) -ldflags $(GO_LDFLAGS)
 GOLANGCI_LINT := golangci-lint
+ifneq ($(GOARM),)
+	ZREPL_TARGET_TUPLE := $(GOOS)-$(GOARCH)v$(GOARM)
+else
+	ZREPL_TARGET_TUPLE := $(GOOS)-$(GOARCH)
+endif
 
 .PHONY: printvars
 printvars:
 	@echo GOOS=$(GOOS)
 	@echo GOARCH=$(GOARCH)
+	@echo GOARM=$(GOARM)
 
 
 ##################### PRODUCING A RELEASE #############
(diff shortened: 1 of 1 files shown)


MAFLO321:
@problame should we also support armel (armv6)?
I think not, because Debian, Ubuntu, Fedora do not support it offially.
If someone in the community asks for it, it can be added easily later on.

Makefile
```diff
@@ -94,11 +96,17 @@ clean: docs-clean
 
 BINS_ALL_TARGETS := zrepl-bin platformtest-bin vet lint
 GO_SUPPORTS_ILLUMOS := $(shell $(GO) version | gawk -F '.' '/^go version /{split($$0, comps, " "); split(comps[3], v, "."); if (v[1] == "go1" && v[2] >= 13) { print "illumos"; } else { print "noillumos"; }}')
+ifeq ($(GOARM), 7)
+	PLATFORM := $(GOOS)-$(GOARCH)hf
+else
+	PLATFORM := $(GOOS)-$(GOARCH)
+endif
```
problame: empty line after `endif` please

Makefile
```diff
@@ -94,11 +96,17 @@ clean: docs-clean
 
 BINS_ALL_TARGETS := zrepl-bin platformtest-bin vet lint
 GO_SUPPORTS_ILLUMOS := $(shell $(GO) version | gawk -F '.' '/^go version /{split($$0, comps, " "); split(comps[3], v, "."); if (v[1] == "go1" && v[2] >= 13) { print "illumos"; } else { print "noillumos"; }}')
+ifeq ($(GOARM), 7)
+	PLATFORM := $(GOOS)-$(GOARCH)hf
```
problame: I'd rather encode the `$GOARM` value as `$(GOARCH)v$(GOARM)`.
Also, please rename `PLATFORM` to `ZREPL_TARGET_TUPLE` to avoid confusion with the platform tests, and move the definition to the top of the file (i.e. right before `printvars`).

problame:
> @problame should we also support armel (armv6)?
> I think not, because Debian, Ubuntu, Fedora do not support it offially.
> If someone in the community asks for it, it can be added easily later on.

No, I think ARMv7 is fine for the time being, I don't want to make the CI jobs run longer than necessary.

Makefile
MAFLO321: I'm not shure what the right arch name is `armhf`, `armv7`, `armv7hl`:
Debian: apache2_2.4.38-3+deb10u3_armhf.deb
Ubuntu: apache2_2.4.41-1ubuntu1_armhf.deb
Fedora: httpd-2.4.41-13.fc32.armv7hl.rpm
FreeBSD: ports branch armv7
OpenSUSE: armv7hl
Arch Linux ARM: apache-2.4.41-1-armv7h.pkg.tar.xz

MAFLO321: used `$(GOARCH)v$(GOARM)` for now

awehrfritz:
Just to comment on the ARMv6 architecture: I don't think there is much need for that. 

People running ZFS on ARM systems typically use SoCs with at least ARMv7 CPUs (ARMv6 would be 6-7 year old hardware such as a RPi 1). The [Helios4](https://wiki.kobol.io/helios4/intro/) is a very attractive platform for a DIY NAS appliance due to the low-power CPU, 4 native SATA ports and ECC memory.

PR #281, opened 2020-02-20
Fix typos in jobid.go

InsanePrawn:

### DIFF
--- a/endpoint/jobid.go
+++ b/endpoint/jobid.go
@@ -9,8 +9,8 @@ import (
 	"github.com/zrepl/zrepl/zfs"
 )
 
-// An instance of this type returned by MakeJobID guarantees
-// that that instance's JobID.String() can be used in a ZFS dataset name and hold tag.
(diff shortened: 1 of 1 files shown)


endpoint/jobid.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/zrepl/zrepl/zfs"
 )
 
-// An instance of this type returned by MakeJobID guarantees
-// that that instance's JobID.String() can be used in a ZFS dataset name and hold tag.
+// JobID instances returned by MakeJobID() guarantee JobID.String() to
+// return a string that can be used in a ZFS dataset name and hold tag.
```
problame: `guarantee that JobID.String() will return a string` ...?

problame:
please prefix the commit message with `endpoint: ...`

endpoint/jobid.go
InsanePrawn: Gave it another shot. If you dislike the revision I'll revert to the old "that that'.

PR #284, opened 2020-02-25
platformtest: run gofmt on replicationCursor.go

InsanePrawn:
`make format` reeeeally wants to add that newline 🤷‍♂️


problame:
curious, why doesn't the CI complain about this?

- [ ] please change the commit message to include the term 'platformtest'
- [ ] could you investigate whether there's anything broken in the way the CI installs dependencies (`.circleci/config.yml`) vs. we tell devs to do it (`lazy.sh godep`)?

problame:
This is an empty patch when rebased onto current master.

PR #285, opened 2020-02-25
zfs send/recv arguments: properties (-p, -b; -o, -x) , on-disk format (raw / -Lce), --saved

### LINKED ISSUE #24: Property Replication
Right now, all `zfs send` invocations on the sending side are without `-p`, meaning we do not replicate any properties.

On the receiving side, as a safeguard, we override `mountpoint` in order to protect ourselves from a malicious `sender` that is trying to mount over some filesystem on the backup server.

This is more of a theoretical advantage and I'm investigating better solutions, e.g. `zfs receive -x all`, see https://github.com/zrepl/zrepl/wiki/ZFS-Feature-Support-&-Wishlist#zfs-receive--x-all

## Agenda
- [ ] Find a way how property replication should work
  - Where to store properties on receiver?
  - Protect against malicious sender (`zfs receive -x all`)
  - Think about restore procedures
- [ ] Implement it
- [ ] Document it (refs #23 )

### LINKED ISSUE #276: Support unencrypted raw sends
I guess this borders on #182. 

As discussed in https://github.com/zrepl/zrepl/pull/259#issuecomment-584615614, I think we need the ability to use `send -w` on unencrypted datasets, if only for `-L` to circumvent zfsonlinux/zfs#6224, but it's also useful for skipping de- and recompression, amongst other things.

As I said there:
I suggest splitting up the two features: an option for raw sends, and an option for enforcing encryption.
Encryption implies raw, raw doesn't imply encryption
=> encryption=on + raw=off is a config error, encryption=(off|unset) + raw=(on|off|unset) is fine.
encryption=off + raw=off is the default anyway.

(If whoever implements this feels super nice, one could check whether `send` supports `-w` and substitute with `-Lce` for unencrypted sends, but I consider this very optional. All these flags are available in the now slowly aging ZoL 0.7 series. If you're running 0.6 in the year 2020, just don't turn 'raw' on in zrepl I guess; oooor we could go the extra mile and try to check for those feature flags being enabled on the pool and passing the enabled subset to `send`.)

InsanePrawn:
Adds config values:
- for sending dataset properties on the sending side (`-p`)
- for sending the received value of properties rather than the local value when available (`-b`)
- for inheriting (`-x`) or overriding (`-o`) properties with fixed values on the receiving side
- sending data as it is on disk: Large blocks (`-L`), compressed (`-c`), embedded (`-e`)
- raw (`-w`) sends without enforced encryption; falls back to `-Lce` if `-w` unavailable

Closes #276
Closes #24 

-------

Syntax

```yaml
job:
  send:
    raw: true
    send_properties: true
    backup_properties: true
```

```yaml
job:
  recv:
    properties:
      override:
        mountpoint: none
      inherit:
        - compression
```

ToDo:
- [ ] Docs
- [ ] Sample conf
- [ ] Tests


endpoint/endpoint.go
```diff
@@ -418,6 +423,19 @@ func (c *ReceiverConfig) copyIn() {
 
 func (c *ReceiverConfig) Validate() error {
 	c.JobID.MustValidate()
+
+	if c.InheritProperties != nil {
+		for _, prop := range c.InheritProperties {
+			prop.MustValidate()
```
problame: should not panic but return an error instead, same below

endpoint/endpoint.go
```diff
@@ -680,7 +698,8 @@ func (s *Receiver) Receive(ctx context.Context, req *pdu.ReceiveReq, receive zfs
 
 	// determine whether we need to rollback the filesystem / change its placeholder state
 	var clearPlaceholderProperty bool
-	var recvOpts zfs.RecvOptions
+	recvOpts := zfs.RecvOptions{InheritProperties: s.conf.InheritProperties, OverrideProperties: s.conf.OverrideProperties}
```
problame: make this multiline please

zfs/zfs.go
```diff
@@ -1032,6 +1037,34 @@ type RecvOptions struct {
...
+	InheritProperties    []Property
+	OverrideProperties   map[Property]string
+}
+
+type Property string
+
+func (p Property) Validate() error {
+	// see zfsprops(8), section "User Properties"
+	if len(p) < 1 {
+		return fmt.Errorf("property name can't be empty")
```
problame: s/can't/cannot/

zfs/zfs.go
```diff
@@ -1032,6 +1037,34 @@ type RecvOptions struct {
...
+	}
+	if len(p) > 256 {
```
problame: please find the ZFS constant, extract it to a const variable and put a comment over it. also make sure that C max string lengths often include the `\0` character

zfs/zfs.go
```diff
@@ -1032,6 +1037,34 @@ type RecvOptions struct {
...
+		return fmt.Errorf("property name longer than 256 characters")
+	}
+	if p[0] == '-' {
+		return fmt.Errorf("property name can't start with '-'")
```
problame: - [ ] `can't`
- [ ] this seems fairly arbitrary, please put a comment here that explains why we only need to check for `-`

zfs/zfs.go
```diff
@@ -1032,6 +1037,34 @@ type RecvOptions struct {
...
+	}
+	propertyValidChar := regexp.MustCompile(`^[0-9a-zA-Z-_\.:]+$`)
```
problame: please extract this to a private global variable

also: where did this come from?

zfs/zfs.go
```diff
@@ -1032,6 +1037,34 @@ type RecvOptions struct {
...
+	if !propertyValidChar.MatchString(string(p)) {
+		return fmt.Errorf("property name must only contain alphanumeric chars and any in %q", "-_.:")
```
problame: you can `.String()` a regex

zfs/zfs.go
```diff
@@ -1111,6 +1144,17 @@ func ZFSRecv(ctx context.Context, fs string, v *ZFSSendArgVersion, streamCopier
 		}
 		args = append(args, "-s")
 	}
+	if opts.InheritProperties != nil {
+		for _, prop := range opts.InheritProperties {
+			args = append(args, "-x", string(prop))
```
problame: FreeBSD 12 does not support `-x` nor `-o` ( https://www.freebsd.org/cgi/man.cgi?zfs(8) )

Thus we'll need feature detection for this, similar to how we do for resumable send ( https://github.com/zrepl/zrepl/blob/44bd354eaedafd7eb5e92e4cf8f7b896e996459a/zfs/resume_token.go#L42 )

zfs/zfs.go
```diff
@@ -1111,6 +1144,17 @@ func ZFSRecv(ctx context.Context, fs string, v *ZFSSendArgVersion, streamCopier
...
+		}
+	}
+	if opts.OverrideProperties != nil {
+		for prop, value := range opts.OverrideProperties {
+			args = append(args, "-o", fmt.Sprintf("%s=%s", prop, value))
```
problame: see above

zfs/zfs.go
```diff
@@ -328,6 +328,10 @@ func (a ZFSSendArgs) buildCommonSendArgs() ([]string, error) {
 		args = append(args, "-w")
 	}
 
+	if a.Properties.B {
+		args = append(args, "-p")
```
problame: I guess we're good wrt compatibility with old ZFS versions here (clear from FreeBSD side, I guess ZoL already had this in 0.7, too?)

endpoint/endpoint.go
```diff
@@ -418,6 +423,19 @@ func (c *ReceiverConfig) copyIn() {
...
+
+	if c.InheritProperties != nil {
+		for _, prop := range c.InheritProperties {
+			prop.MustValidate()
+		}
+	}
+
+	if c.OverrideProperties != nil {
+		for prop, _ := range c.OverrideProperties {
+			prop.MustValidate()
```
problame: see above

endpoint/endpoint.go
```diff
@@ -408,8 +412,9 @@ type ReceiverConfig struct {
 
 	RootWithoutClientComponent *zfs.DatasetPath // TODO use
 	AppendClientIdentity       bool
-
-	UpdateLastReceivedHold bool
+	InheritProperties          []zfs.Property
+	OverrideProperties         map[zfs.Property]string
```
problame: we'll need to `ReceiverConfig.copyIn` both `InheritProperties` and `OverrideProperties` (this is a method right below this one, please put a comment in here to make this sufficiently clear for future modifications)

endpoint/endpoint.go
```diff
@@ -418,6 +423,19 @@ func (c *ReceiverConfig) copyIn() {
 
 func (c *ReceiverConfig) Validate() error {
 	c.JobID.MustValidate()
```
problame: you might have used `MustValidate` below because of this call to `JobID.MustValidate`.

I dont' remember why I made those call sites not just use `JobID.Validate`.

However, a `Validate` function like this one should always error and not panic.

zfs/zfs.go
InsanePrawn: This is an error message i copy-paste stole from you :) I don't think "doesn't match regex: [ascii puke]" is help- or meaningful to most users. The current message is somewhat more 'outspoken' than `^[0-9a-zA-Z-_\.:]+$`.

zfs/zfs.go
InsanePrawn: Literally from the above mentioned man page section. I've moved the comment above the function so it's clearer it applies to the whole function.

zfs/zfs.go
InsanePrawn: man page strikes again :)

endpoint/endpoint.go
InsanePrawn: I'm unsure why we need to do that, although I can imagine a few possible reasons. Do you have a  suggestion for such a comment?

InsanePrawn:
> Thanks for this PR!

Thanks for the quick review!

> Some other remarks:
> 
> - [ ] Consider validating the special-cases mentioned in the man page:

Ugh. So either we stay out of this, delegating proper property handling to the user completely, or we have to try to guess what the user needs. Maybe we can have some basic profiles, here's my suggestions:
- `send_properties: true` implies `-o mountpoint=none`
- `encrypted: true` implies `-x keylocation` or some other encroot workaround, TBD.

Do we make these always on? Default to on but can be disabled?

> * [ ]  What about transformations of `mountpoint`?
>     
> * the default override should be `mountpoint=none` IMHO, but I'm open to discussion
> * users will surely want to put a **prefix** on the received mountpoint property

I was careful to stress the 'static' and 'global' part of property overrides. I will not implement dynamically rewriting properties. I consider this optional future work.
 
>     * [ ]  Consider throwing in the send option for `--backup` as well. Might prove useful in your multi-stage replication setup ;)

I'll add feature detection and `-b` in a new revision

>       * AFAIK receiving with `mountpoint` != none modifies the filesystem into which we mount into, prohibiting future receives to it until someone rolls it back manually. Is that actually a problem? What to do about that?

I can definitely say `mountpoint=none` works as expected.
I'll test other values the next time I work on this; but IMHO just because you _can_ specify `-o mountpoint=/something/not/none` doesn't mean you should or that it has to work. `-o gun:aim=foot`, if you will.
 I definitely don't expect `zrepl configcheck` to catch that and I'd rather keep zrepl's semantic understanding of specific properties as non-existent as possible.

config/config_recv_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	recv_empty := `
+  recv: {}
+`
+
+	recv_not_specified := `
+`
+
+	fill := func(s string) string { return fmt.Sprintf(tmpl, s) }
+	var c *Config
```
problame: shouldn't Go be complaining about this unused variable? (I think the sub-tests shadow `c` when they call `c, err := testConfig(...` )

endpoint/endpoint.go
problame: we do the copy-in to ensure that the `ReceiverConfig` object is self-contained.

I guess the comment should be on the entire `ReceiverConfig` struct:

```
+ // NOTE: when adding members to this struct, remember
+ //       to add them to `ReceiverConfig.copyIn()`
  type ReceiverConfig struct {
```

zfs/zfs.go
problame: agreed, but then either put comments at the regex definition and fmt.Errof sites, or move the fmt.Errorf closer to the `propertyValidNameChars` definition like so:

```
// Keep regex and error message in sync!
var (
  propertyValidNameChars = regexp.MustCompile(`^[0-9a-zA-Z-_\.:]+$`)
  propertyValidNameCharsErr = fmt.Errorf("property name must only contain alphanumeric chars and any in %q", "-_.:")
)
```

endpoint/endpoint.go
```diff
@@ -408,16 +412,46 @@ type ReceiverConfig struct {
...
 }
 
 func (c *ReceiverConfig) Validate() error {
 	c.JobID.MustValidate()
+
+	if c.InheritProperties != nil {
+		for _, prop := range c.InheritProperties {
+			err := prop.Validate()
+			if err != nil {
+				return err
```
problame: the returned error message doesn't contain the offending property yet.
```
errors.Wrapf(err, "inherit property %q", prop)
```

endpoint/endpoint.go
```diff
@@ -408,16 +412,46 @@ type ReceiverConfig struct {
...
+			}
+		}
+	}
+
+	if c.OverrideProperties != nil {
+		for prop := range c.OverrideProperties {
+			err := prop.Validate()
+			if err != nil {
+				return err
```
problame: Same here:
```
errors.Wrapf(err, "override property %q", prop)
```

zfs/zfs.go
```diff
@@ -618,11 +638,21 @@ func (n *NilBool) String() string {
 	return fmt.Sprintf("%v", n.B)
 }
 
+func (n *NilBool) IsTrue() bool {
```
problame: Sry, can't have that!
The reason for `NilBool`'s existence is to have a type that doesn't default to something:

```
type Foo struct {
  ImportantConfigKnob *NilBool
}

...

func (f Foo) {
   if f.ImportantConfigKnob.B {
   
   } else {
   }
}
```

If someone declares such a struct, they express the intent that they always want it be set explicitly, and that they would rather crash than have the implicit default value there (because the caller forgot to set it explicitly).
(Default values are a double-edged sword, and for important features such as *should we send the user data encrypted or unencrypted*, I'd much rather crash if we forgot to set that field somewhere, than do the wrong thing).

---

Feel free to add the above as a doc comment on `NilBool`.

---

As to your usage of `NilBool`: I think most of the send options (`LargeBlock`, `EmbeddedData`, etc. can safely default to `false`). Thus you could just use a normal `bool`. But if you think the caller should be punished for forgetting to set the fields, leave them as `NilBool`s. You just can't have `IsTrue()`, it defeats the purpose of `NilBool`.

zfs/zfs.go
```diff
@@ -814,14 +844,28 @@ func ZFSSend(ctx context.Context, sendArgs ZFSSendArgs) (*ReadCloserCopier, erro
...
+				return nil, ErrEncryptedSendNotSupported
+			}
+
+			if sendArgs.Raw.IsTrue() {
+				// zfs-send(8) about `send -w`: "For unencrypted datasets, this flag will be equivalent to -Lec."
+				// TODO: expose via config?
+				// TODO: definitely add feature checks (also for -b!)
+				sendArgs.LargeBlocks = &NilBool{true}
+				sendArgs.EmbeddedData = &NilBool{true}
+				sendArgs.Compressed = &NilBool{true}
```
problame: `sendArgs` should not be mutated in `ZFSSend`: it expresses intent of the caller. I can't seem to find the place where you convert the `config.SendOptions.Raw` to a `ZFSSendArgs.Raw`. Wherever that is / will be, that's where above smartness should be. Plus some unit tests for that smartness ;)

zfs/zfs.go
InsanePrawn: Ah, thanks! While writing this I was already wondering whether this is the right thing to do.

config/config.go
```diff
@@ -82,12 +82,15 @@ type SendOptions struct {
...
+	LargeBlocks      bool `yaml:"large_blocks,optional"`
+	Compressed       bool `yaml:"compressed,optional"`
+	EmbeddedData     bool `yaml:"embbeded_data,optional"`
 }
 
 var _ yaml.Defaulter = (*SendOptions)(nil)
 
 func (l *SendOptions) SetDefault() {
-	*l = SendOptions{Encrypted: false, Raw: false, SendProperties: false, BackupProperties: false}
+	*l = SendOptions{Encrypted: false, Raw: false, SendProperties: false, BackupProperties: false, LargeBlocks: false, Compressed: false, EmbeddedData: false}
```
problame: please switch to one field per line here

zfs/zfs.go
InsanePrawn: Yup, IRC assured me back then that this is not something new if you're at least on 0.7.x IIRC. As this isn't on by default, I'm against bloating this with a compatibility check.

gbarazer:
Hi, I just saw this contribution and as you are considering adding possible arguments for zfs send/recv (which is a very interesting feature), would it be possible to consider adding zfs send -c and zfs recv -d and -e as well ? Our use case is 1:1 replication between to node for HA / failover purpose and without -d, zrepl is simply unusable in this use case.

InsanePrawn:
Hi, thanks for your interest!

>  would it be possible to consider adding zfs send -c

Toggles for send -L, -c, -e and -w are included in the PR already.

> and zfs recv -d and -e as well ? Our use case is 1:1 replication between to node for HA / failover purpose and without -d, zrepl is simply unusable in this use case.

Reading the docs for these, I'm unsure what they do for you. Can you describe your setup in more detail and how the flags play into it? We're interested in making zrepl as flexible as possible for as many use cases as possible [to some degree].
Have you reviewed the exact zfs send and receive calls zrepl makes? Are they compatible with your plans [given that recv -e and -d get implemented]? Please note that zrepl uses neither `zfs snap -r` nor `zfs send -R`.

> I also agree that perhaps we should just let the use[r] decide to add any arguments to the zfs send and recv commands without zrepl trying to understand the arguments, which will increasingly complexify its code and usage. We could add first raw arguments, then gradually more logic related to the passed arguments as new features to better handle the logic end-to-end (like let zfs send -c passed with raw arguments, then add later a more complex logic to let the user specify which compression algorithm she wants and handle -c transparently). I think this would greatly increase the adoption rate for zrepl.

Do you have more examples of flags not in this PR that people might want to stick into their arg arrays and that would work with zrepl? I'm having a hard time coming up with any.
FWIW with the raw send (-w) in this PR, we shouldn't need new parameters for new on-disk features any time soon, hopefully.

gbarazer:
Hi,

The case for zfs recv -d is very simple, we want to replicate a full pool from one HA node to the other (which is not exactly HA, because we have a 1-2 minute delay between both nodes due to replication, but allow us to have both nodes with separate pools able to serve NFS requests and failover very quickly without having to re-import a pool).
The current zrepl logic always tries to receive a pool into a predefined destination (which forbids us the use of "sink" and a push logic), and when pulling datasets, it always tries to recv with the source pool name into our destination.
Example:
source node with pool sourcepool/*
dest node with pool destpool
pull job always creates destpool/sourcepool/*  and there is no way to receive directly sourcepool/* into destpool/*

I agree that instead of using the -d flag we could change the logic implemented into the "sink" or "pull" jobs to have the same behaviour, but it's currently not possible to replicate a full pool without having the pool name into the destination.

problame:
I moved the discussion for the 1:1 pool replication use case [to a new issue](https://github.com/zrepl/zrepl/issues/330) as it is not related to the per-filesystem flags that this PR will bring.

jwittlincohen:
What would be the best way to get large block support now currently with zrepl?  

I really like the robustness and ease-of-configuration of zrepl, but unfortunately, the lack of large block support is a major problem for my use case.  My main pool has two 12-disk raidz2 vdevs.  With 1M blocks, parity+padding overhead is 17%, but with 128k, it's 24%.  This difference amounts to almost a year of newly written data.  Unlike with compression, once the blocks are written at a smaller recordsize, there's no way to get back the original 1M blocks with a zfs send.  In other words, even if the receiving dataset has recordsize=1M, if the blocks are 128k, they'll be sent as 128k blocks.  As a result, I would need to use rsync or an equivalent tool, which would take 2-3x longer than a zfs send over 10Gb, and would result in losing my snapshots.  Moreover, I have found rsync to generally require more manual intervention due to various permission issues.  This could easily be resolved if I could just specify -L on the sending side.   Large block support was added in 0.6.5, which released in September 2015, so I doubt many people are running systems without support for this feature.  

So, is there any easy way for me to implement large-block support currently, even if it is hacky? In addition, are you accepting bounties to push forward the release of new features? I would be happy to contribute to such a bounty to get such a feature implemented more quickly.  I apologize if this is the wrong venue to ask these questions.  I wasn't sure where else to ask.

problame:
> What would be the best way to get large block support currently in zrepl?
> So, is there any easy way for me to implement large-block support currently, even if it is hacky?

Write a wrapper `zrepl` shell script that you put in the `PATH` of the zrepl daemon.
The wrapper shell script intercepts the `zfs send` invocations and adds the `-L` flag.
An example for this is a script we use in the platformtests: https://github.com/zrepl/zrepl/blob/master/platformtest/logmockzfs/zfs

Alternatively, if you are willing to patch zrepl, it's quite a simple thing to add the `"-L"` to the arguments slice in the place where we compose the zfs command line for `ZFSSend`:

https://github.com/zrepl/zrepl/blob/83fdffbceff9d1c8b1690dc104f9cdbb7b2f8b87/zfs/zfs.go#L322-L355

> In addition, are you accepting bounties to push forward the release of new features? I would be happy to contribute to such a bounty to get such a feature implemented more quickly. I apologize if this is the wrong venue to ask these questions. I wasn't sure where else to ask.

Such a  bug / feature bounty would be the first one in zrepl's 'history'. I set up a BountySource-account more than a year ago.
However, since @InsanePrawn started this PR, I would like to know what his schedule for this PR looks like.

jwittlincohen:
> Alternatively, if you are willing to patch zrepl, it's quite a simple thing to add the `"-L"` to the arguments slice in the place where we compose the zfs command line for `ZFSSend`:
> 
> https://github.com/zrepl/zrepl/blob/83fdffbceff9d1c8b1690dc104f9cdbb7b2f8b87/zfs/zfs.go#L322-L355
> 

How would this interact with appending -w for use with encrypted datasets? It seems like this would result in -wL for encrypted sends, and I'm not sure whether zfs will accept -L and -w together.

> > In addition, are you accepting bounties to push forward the release of new features? I would be happy to contribute to such a bounty to get such a feature implemented more quickly. I apologize if this is the wrong venue to ask these questions. I wasn't sure where else to ask.
> 
> Such a bug / feature bounty would be the first one in zrepl's 'history'. I set up a BountySource-account more than a year ago.
> However, since @InsanePrawn started this PR, I would like to know what his schedule for this PR looks like.

In that case, what would be the preferred method to support the projects? Monthly Patreon contribution? Single larger contribution through Paypal?

problame:
> How would this interact with appending -w for use with encrypted datasets? It seems like this would result in -wL for encrypted sends, and I'm not sure whether zfs will accept -L and -w together.
> If I'm patching to add -L, is there any reason not to add -Lce?
> EDIT: Nevermind, I just tested this and the send/recv worked as expected. It appears to just ignore the -L,

Great, maybe post your patch here so people can use it until this PR lands!

> In that case, what would be the preferred method to support the project? Monthly Patreon contribution? Single larger contribution through Paypal?

GitHub Sponsors doubles the donation ATM, so that's probably the best way to go.
A big PayPal donation would certainly be nice as well, but Patreon is just as good and probably increases zrepl's publicly perceived "relevance" a little.
Note that zrepl will remain a spare-time project for the time being, and this specific PR is currently not managed by myself but @InsanePrawn.
Disclaimer: At this time, all advertised ways to support the project yield the money to me, solely. I invest the money into OpenZFS conference trips.

In any way, thanks for supporting zrepl!

jwittlincohen:
I'm attempting to build zrepl using your docker instructions (installed Docker on Debian Buster using the official instructions from https://docs.docker.com/engine/install/debian/) but am hitting failures on `sudo docker run -it --rm \
    -v "${PWD}:/src" \
    --user "$(id -u):$(id -g)" \
    zrepl_build make release` 

Here's the console output: https://pastebin.com/fJPWsDim

Any idea why it's failing?

problame:
> I'm attempting to build zrepl using your docker instructions (installed Docker on Debian Buster using the official instructions from https://docs.docker.com/engine/install/debian/) but am hitting failures on sudo docker run -it --rm \ -v "${PWD}:/src" \ --user "$(id -u):$(id -g)" \ zrepl_build make release
> 
> Here's the console output: https://pastebin.com/fJPWsDim
> 
> Any idea why it's failing?

Yes, it's because the Docker container is `FROM golang:latest` which got bumped to Go 1.15 which was released yesterday.

I have the fixes ready in branch `problame/docs-update-and-fix-go1.15`, so **just rebase your patches onto that branch**.
(There are some [ARM build failures](https://app.circleci.com/pipelines/github/zrepl/zrepl/662/workflows/05dbe6ed-437c-4d10-a47f-17914d01d4f6/jobs/2852) that are blocking that branch from getting merged).

problame:
> EDIT: I enrolled in the monthly sponsorship on Github. Please alert me if, for whatever reason, you don't receive the payment.

Thank you very much, as you might have noted, you are in supporters.rst already, will show up on https://zrepl.github.io/supporters.html when I release 0.3 (or the next 0.3-rc)

jwittlincohen:
I manually applied the changes from your Go 1.15 branch and was able to build the binaries in Docker.  I modified lines 351-353 of `zrepl/zfs/zfs.go ` to the following:

```
                args = append(args, "-Lce", toV)
        } else {
                args = append(args, "-Lcei", fromV, toV)
```

Since I didn't know how to directly observe the zfs send command used by zrepl, I did some indirect tests.  I created a new dataset on the source side called `data/test` initially set to `recordsize=1M` and `compression=off.`  I generated a 1GB file using dd to write zeroes (dd if=/dev/zero of=blob_no_compression bs=1M count=1000).  I then set "compression=lz4" on the dataset and created a new 1GB file.  Finally, I created a 1GB file using pseudo-random data from /dev/urandom.  I then setup zrepl to replicate this to my backup pool where recordsize=1M and compression=lz4.  I observed the following:

- The replication showed a 2GB transfer, indicating that the 1GB file of zeroes generated when compression=lz4 was sent using compressed blocks. zdb confirms that this object has a 0 dsize.

```
    Object  lvl   iblk   dblk  dsize  dnsize  lsize   %full  type
       256    1   128K     1M      0      1K     1M    0.00  ZFS plain file
                                               176   bonus  System attributes
        dnode flags: USED_BYTES USERUSED_ACCOUNTED USEROBJUSED_ACCOUNTED
        dnode maxblkid: 0
        path    /blob_compression
```

- zdb -ddddd showed that all 3 files had a dblk size of 1M.  An inspection of the individual blocks on the encrypted 1GB file indicated, as expected, that both the logical and physical block size was 1M.  For example, "          600000  L0 0:674a9a4f8000:135000 100000L/100000P F=1 B=331187/331187 cksum=20052a991e3da:fffbf384bf9ca4bb:416b28d8266fb6ff:a9c29fb7d580d969"

- Because zrepl does not send properties, the dataset inherited compression=lz4 from its parent dataset.  As such, the 1GB file generated on the source side with compression=off, while sent in its original, uncompressed form, was compressed upon writing to the pool.

`    Object  lvl   iblk   dblk  dsize  dnsize  lsize   %full  type
       256    1   128K     1M      0      1K     1M    0.00  ZFS plain file
                                               176   bonus  System attributes
        dnode flags: USED_BYTES USERUSED_ACCOUNTED USEROBJUSED_ACCOUNTED
        dnode maxblkid: 0
        path    /blob_compression
`

config/config_recv_test.go
InsanePrawn: Actually the mix of local and global assignments kept it used.

I've converted everything to sub-test local variables as we don't seem to use `c` outside.

endpoint/endpoint.go
```diff
@@ -169,15 +191,37 @@ func (s *Sender) Send(ctx context.Context, r *pdu.SendReq) (*pdu.SendRes, io.Rea
...
 		return nil, nil, errors.Wrap(err, "validate send arguments")
 	}
 
+	if s.sendRaw {
+		encryptionSupported, err := zfs.EncryptionCLISupported(ctx)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "cannot determine CLI native encryption support")
+		}
+
+		if !encryptionSupported {
```
InsanePrawn, line 216: TODO: remove this `-w -> -Lce`  logic and just let it fail, the user can always configure -Lce instead

config/config_recv_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	t.Run("recv_override_empty", func(t *testing.T) {
+		c := testValidConfig(t, fill(recv_override_empty))
+		assert.NotNil(t, c)
+	})
+
+	t.Run("recv_override", func(t *testing.T) {
+		c := testValidConfig(t, fill(recv_override))
+		override := c.Jobs[0].Ret.(*PullJob).Recv.Properties.Override
+		assert.NotEmpty(t, override)
```
problame, line 100: you should probably test for `testprop2: "test123"` as well, similarly in the other test cases

config/config_send_test.go
```diff
@@ -38,36 +38,108 @@ jobs:
...
+	})
+
+	t.Run("raw_false", func(t *testing.T) {
+		c := testValidConfig(t, fill(raw_false))
+		raw := c.Jobs[0].Ret.(*PushJob).Send.Raw
+		assert.Equal(t, false, raw)
+	})
+
+	t.Run("raw_and_encrypted", func(t *testing.T) {
+		c := testValidConfig(t, fill(raw_and_encrypted))
```
problame, line 134: I think `raw` and `encrypted` should be mutually exclusive in the config.
Logically, they can be combined, but since `send.encrypted=true` will already enforce that all filesystems matched by `filesystems` are encrypted, `send.raw=true` has no effect.

endpoint/endpoint.go
problame: Yeah, ISTM this whole `s.sendRaw` block should go away.

zfs/zfs.go
```diff
@@ -329,10 +330,30 @@ func (a ZFSSendArgsUnvalidated) buildCommonSendArgs() ([]string, error) {
 		return args, nil
 	}
 
-	if a.Encrypted.B {
+	if (a.Encrypted != nil && a.Encrypted.B) || a.Raw {
```
problame, line 333: a.Encrypted must always be not nil, it's a logic error in caller if` a.Encrypted == nil`.
I'll push a commit to this branch that explains the `NilBool` in a doc comment.

zfs/zfs.go
```diff
@@ -779,17 +807,25 @@ func ZFSSend(ctx context.Context, sendArgs ZFSSendArgsValidated) (*SendStream, e
...
 			return nil, errors.Wrap(err, "cannot determine CLI native encryption support")
 		}
-		if !supported {
+
+		if !encryptionSupported {
 			return nil, ErrEncryptedSendNotSupported
 		}
 	}
 
+	if _, err := sendArgs.Validate(ctx); err != nil {
```
problame, line 825: This doesn't make sense sinse sendArgs are already `ZFSSendArgsValidated`.

Was this supposed to be the `sendArgs.Encrypted.Validate()` call? (see left side of this diff)

zfs/zfs.go
```diff
@@ -1006,6 +1034,40 @@ func (e *ErrRecvResumeNotSupported) Error() string {
...
+	// Note that this doesn't change property values, i.e. an existing local property value will be kept.
+	RollbackAndForceRecv bool
+	// Set -s flag used for resumable send & recv
+	SavePartialRecvState bool
+
+	InheritProperties  []zfsprop.Property
+	OverrideProperties map[zfsprop.Property]string
+}
+
+func (opts RecvOptions) buildCommonRecvArgs() []string {
```
problame, line 1048: Why is this method called `buildCommonRecvArgs` and not just `buildRecvArgs`? The method is used in the same way from all call sites, the call sites don't extend the returned argv with more flags

problame (changes requested):

zfs/zfs.go
InsanePrawn: Maybe, I kinda questioned myself about this chunk while rebasing.
Question though: Why do we revalidate Encrypted (in the original code) here? It's already validated in `ZFSSendArgsUnvalidated.Validate()`.
The comment mentions go1.13? Is this an oversight?

zfs/zfs.go
InsanePrawn: Funnily enough I named it so because I borrowed the naming from buildCommonSendArgs, which is used as you describe and includes specific things like snapshots and dataset names, while buildCommonRecvArgs actually does only build the 'common' args, as in flags from booleans.
For some of the testing I added it would be nice to split SendArgsUnvalidated.buildCommonSendArgs() into SendArgsValidated.buildSendArgs() (as suggested), which gets its flags from SendArgsUnvalidated.buildCommonSendArgs().
Not sure if this is beneficial beyond being able to test it.

zfs/zfs_test.go
```diff
@@ -304,3 +306,55 @@ func TestTryRecvDestroyOrOverwriteEncryptedErr(t *testing.T) {
...
+		"Empty Args":                 {conf: RecvOptions{}, argsInclude: []string{}, argsExclude: []string{"-x", "-o", "-F", "-s"}},
+		"ForceRollback":              {conf: RecvOptions{RollbackAndForceRecv: true}, argsInclude: []string{"-F"}, argsExclude: []string{"-x", "-o", "-s"}},
+		"PartialSend":                {conf: RecvOptions{SavePartialRecvState: true}, argsInclude: []string{"-s"}, argsExclude: []string{"-x", "-o", "-F"}},
+		"Override properties":        {conf: RecvOptions{OverrideProperties: map[zfsprop.Property]string{zfsprop.Property("abc"): "123"}}, argsInclude: []string{"-o", "abc=23"}, argsExclude: []string{"-x", "-F", "-s"}},
+		"Exclude/inherit properties": {conf: RecvOptions{InheritProperties: []zfsprop.Property{"abc", "123"}}, argsInclude: []string{"-x", "abc", "123"}, argsExclude: []string{"-o", "-F", "-s"}},
+	}
+
+	for testName, test := range recvTests {
+		t.Run(testName, func(t *testing.T) {
+			args := test.conf.buildCommonRecvArgs() //fill(encrypted_false))
```
InsanePrawn, line 353: leftover comment

zfs/zfs_test.go
```diff
@@ -304,3 +306,55 @@ func TestTryRecvDestroyOrOverwriteEncryptedErr(t *testing.T) {
...
+		"Encrypt":                {conf: ZFSSendArgsUnvalidated{Encrypted: &NilBool{true}}, argsInclude: []string{"-w"}, argsExclude: []string{}},
+		"Raw":                    {conf: ZFSSendArgsUnvalidated{Raw: true}, argsInclude: []string{"-w"}, argsExclude: []string{}},
+		"Send properties":        {conf: ZFSSendArgsUnvalidated{Properties: true}, argsInclude: []string{"-p"}, argsExclude: []string{}},
+		"Send backup properties": {conf: ZFSSendArgsUnvalidated{BackupProperties: true}, argsInclude: []string{"-b"}, argsExclude: []string{}},
+		"Send -b and -p":         {conf: ZFSSendArgsUnvalidated{Properties: true, BackupProperties: true}, argsInclude: []string{"-p", "-b"}, argsExclude: []string{}},
+	}
+
+	for testName, test := range sendTests {
+		t.Run(testName, func(t *testing.T) {
+			args, err := test.conf.buildCommonSendArgs() //fill(encrypted_false))
```
InsanePrawn, line 327: leftover comment.
also:
if we don't refactor buildCommonSendArgs, check Encrypted == nil and default to NilBool{false} inside the loop here to save config bloat above.

InsanePrawn:
did the buildSendArgs vs buildCommonSendArgs refactor, rebased on current master and that refactor, accidentally blew your Nilbool doc commit away in the process and can't find it anymore. 🙃

Some minor-ish things are still open, see unresolved feedback.

config/config.go
```diff
@@ -97,6 +106,23 @@ type ReplicationOptionsProtection struct {
...
+type PropertyRecvOptions struct {
+	Inherit  []zfsprop.Property          `yaml:"inherit,optional"`
+	Override map[zfsprop.Property]string `yaml:"override,optional"`
+}
+
+var _ yaml.Defaulter = (*PropertyRecvOptions)(nil)
+
+func (l *PropertyRecvOptions) SetDefault() {
+	//*l = PropertyRecvOptions{}
+	//TODO: is below necessary?
```
problame, line 122: i think it is, and you should have test cases where `inherit` is set, `override` isn't, and check that the code that consumes PropertyRecvOptions can deal with `Override` being nil.

zfs/zfs.go
```diff
@@ -683,6 +652,76 @@ func (a ZFSSendArgsUnvalidated) Validate(ctx context.Context) (v ZFSSendArgsVali
...
+
+	// ResumeToken takes precedence, we assume that it has been validated
+	// to reflect what is described by the other fields
+	if a.ResumeToken != "" {
+		args = append(args, "-t", a.ResumeToken)
+		return args, nil
+	}
+
+	if a.Encrypted.B && !a.Raw {
+		args = append(args, "-w")
```
problame, line 701: why do we need this here? `ZFSSendArgsUnvalidated.buildCommonSendArgs` should already be setting this, right?

zfs/zfs.go
```diff
@@ -683,6 +652,76 @@ func (a ZFSSendArgsUnvalidated) Validate(ctx context.Context) (v ZFSSendArgsVali
 	}, nil
 }
 
+func (a ZFSSendArgsUnvalidated) buildCommonSendArgs() []string {
+
+	args := make([]string, 0)
+
+	if a.Raw {
```
problame, line 659: what about `.Encrypted.B`?

zfs/zfs_test.go
```diff
@@ -304,3 +306,54 @@ func TestTryRecvDestroyOrOverwriteEncryptedErr(t *testing.T) {
...
+		argsInclude []string
+		argsExclude []string
+	}
+	var sendTests = map[string]SendTest{
+		"Empty Args":             {conf: ZFSSendArgsUnvalidated{}, argsInclude: []string{}, argsExclude: []string{"-w", "-p", "-b"}},
+		"Raw":                    {conf: ZFSSendArgsUnvalidated{Raw: true}, argsInclude: []string{"-w"}, argsExclude: []string{}},
+		"Send properties":        {conf: ZFSSendArgsUnvalidated{Properties: true}, argsInclude: []string{"-p"}, argsExclude: []string{}},
+		"Send backup properties": {conf: ZFSSendArgsUnvalidated{BackupProperties: true}, argsInclude: []string{"-b"}, argsExclude: []string{}},
+		"Send -b and -p":         {conf: ZFSSendArgsUnvalidated{Properties: true, BackupProperties: true}, argsInclude: []string{"-p", "-b"}, argsExclude: []string{}},
+	}
```
problame, line 322: These tests should cover the `.Encrypted` flag as well.

problame (changes requested):
I'll push the NilBool doc comment again.

zfs/zfs.go
InsanePrawn: No, buildCommon works on an unvalidated sendargs struct, and thus only checks raw, not encrypted. But we want to avoid adding `-w` for raw and then again for encrypted, if both are set, which I agree now should error during validation.

TL;DR: yes, we need it here, for the case Raw=false,Encrypted=true.

zfs/zfs.go
```diff
@@ -683,6 +652,76 @@ func (a ZFSSendArgsUnvalidated) Validate(ctx context.Context) (v ZFSSendArgsVali
...
+// a must already be validated
+//
+// SECURITY SENSITIVE because Encrypted must be handled correctly
+func (a ZFSSendArgsValidated) buildSendArgs() ([]string, error) {
+	args := a.buildCommonSendArgs()
+
+	// ResumeToken takes precedence, we assume that it has been validated
+	// to reflect what is described by the other fields
+	if a.ResumeToken != "" {
+		args = append(args, "-t", a.ResumeToken)
```
InsanePrawn, line 725:
```suggestion
		args = []string{"-t", a.ResumeToken}
```

zfs/zfs.go
```diff
@@ -683,6 +652,76 @@ func (a ZFSSendArgsUnvalidated) Validate(ctx context.Context) (v ZFSSendArgsVali
...
+
+	return args
+}
+
+// ResumeToken is allowed to be empty
+// a must already be validated
+//
+// SECURITY SENSITIVE because Encrypted must be handled correctly
+func (a ZFSSendArgsValidated) buildSendArgs() ([]string, error) {
+	args := a.buildCommonSendArgs()
```
InsanePrawn, line 720: restore empty list initialisation and move buildCommonSendArgs below the token resume stuff?

zfs/zfs.go
problame: >  buildCommon works on an unvalidated sendargs struct, and thus only checks raw, not encrypted

How about only ZFSSendArgsValidated can have abuildCommonSendArgs then?

zfs/zfs.go
problame: Yes, if I understand correctly that's what I mean by https://github.com/zrepl/zrepl/pull/285#discussion_r484458572

InsanePrawn:
TODO:
- platformtest: property replication
- docs

problame:
I'm taking over here.

config/config_send_test.go
problame: > I think raw and encrypted should be mutually exclusive in the config.

Only complicates things during testing with no real upside for the user.

config/config_recv_test.go
problame: The integration test covers this end to end.

problame:
* [x] platformtest that exercises property replication
* [x] basic docs
  * could someone please review them?
* [ ] `docs/sendrecvoptions.rst`: improve list of problematic properties
  * @InsanePrawn  @jwittlincohen could you help me out there? just post to this thread...
* [ ] switch over examples to use `{}` notation for `override` section
* [ ] decide on whether we should provide a flag for `zfs send --holds
    * main problem is that it is not possible to exclude holds on the receiving side, so zrepl's holds will always be in the send stream
    * the holds are unique in that they have job ids, but job ids are only unique to a single machine
    * even if we communicate the send-side holds out-of-band, there is no way to atomically recv + release a list of holds

InsanePrawn:
>     * [ ]  `docs/sendrecvoptions.rst`: improve list of problematic properties
>       
>       * @InsanePrawn  @jwittlincohen could you help me out there? just post to this thread...

off the top of my head:
- keylocation
- keyformat?
- encryption
- canmount
- mountpoint
- org.openzfs.systemd:{{wanted,required}-by,ignore}

Symmetrically, these are also properties you may want to proactively set on your sink, except for the crypto stuff.

>     * [ ]  switch over examples to use `{}` notation for `override` section

```yaml
job:
  recv:
    properties:
      override:
        mountpoint: none
        canmount: off
        "org.openzfs.systemd:ignore": on
```
looks okay to me and incidentally "highlights" user props, but I'm not opposed to your change either.

>     * [ ]  decide on whether we should provide a flag for `zfs send --holds
Strong no as we'd need a mechanism to clean them up, at which point we can manually create them as well, which also gives us filtering.
Y tho?

docs/configuration/sendrecvoptions.rst
```diff
@@ -9,37 +9,151 @@ Send & Recv Options
...
+
+.. _job-send-options-large-blocks:
+
+``large_blocks``
+----------------
+
+This flag should not be changed after initial replication.
+Prior to `OpenZFS commit 7bcb7f08 <https://github.com/openzfs/zfs/pull/10383/files#diff-4c1e47568f46fb63546e984943b09a3e6b051e2242649523f7835bbdfe2a9110R337-R342>`_
+it was possible to change this setting which resulted in **data loss on the receiver**.
+The commit in question is included in OpenZFS 2.0 by prohibiting receives of incremental streams with a flipped setting.
```
InsanePrawn, line 91: this sentence seems to be missing a word or two.
```suggestion
The commit in question is included in OpenZFS 2.0 and works around the problem by prohibiting receives of incremental streams with a flipped setting.
```

docs/configuration/sendrecvoptions.rst
```diff
@@ -9,37 +9,151 @@ Send & Recv Options
...
+     recv:
+       properties:
+         inherit:
+           - "mountpoint"
+         override: {
+           "org.openzfs.systemd:ignore": "on"
+         }
+     ...
+
+.. _job-recv-options--inherit-and-override:
```
InsanePrawn, line 117: Is the double dash a typo?

docs/configuration/overview.rst
```diff
@@ -148,6 +148,12 @@ Incremental sends require that ``@from`` be present on the receiving side when r
 Incremental sends can also use a ZFS bookmark as *from* on the sending side (``zfs send -i #bm_from fs@to``), where ``#bm_from`` was created using ``zfs bookmark fs@from fs#bm_from``.
 The receiving side must always have the actual snapshot ``@from``, regardless of whether the sending side uses ``@from`` or a bookmark of it.
 
+**Plain and raw sends**
+ZFS sends the most generic, backwardscompatible data stream format (so-called 'plain send') by default, which means it (amongst other things) decompresses and decrypts data before sending, leaving the receiver with additional recompression and checksumming overhead. Over time, flags like ``-c`` and ``-L`` were added in order to send streams that are closer to how the data is stored on disk, removing computational overhead from senders and receivers. Later, the ``-w`` (``--raw``) flag was added, which summarizes these flags and supports sending data in its encrypted on-disk format.
```
problame, line 152: * Shorter sentences.
* s/like/such as/

docs/configuration/overview.rst
```diff
@@ -148,6 +148,12 @@ Incremental sends require that ``@from`` be present on the receiving side when r
...
+
+.. NOTE::
+    Encrypted datasets can only be sent as either a (unencrypted) fully plain send or a fully raw send (``-w``); specifying single options like ``--compressed`` or ``--larg-block`` is refused for encrypted datasets by zfs send.
```
problame, line 155: remove 'fully

problame: also: --largE-block

docs/configuration/sendrecvoptions.rst
```diff
@@ -79,6 +79,29 @@ If ``encrypted=false``, zrepl expects that filesystems matching ``filesystems``
...
+.. NOTE::
+
+   Please read the :ref:`note on property replication below <job-note-property-replication>`.
+
+.. _job-send-options-backup-properties:
+
+``backup_properties``
+---------------------
+
+When properties are modified on a dataset that was received from a send stream with properties, zfs archives the original received value internally.
```
problame, line 97: - s/dataset/filesystem

docs/configuration/sendrecvoptions.rst
```diff
@@ -79,6 +79,29 @@ If ``encrypted=false``, zrepl expects that filesystems matching ``filesystems``
...
+This also applies to inheriting or overriding properties during receive.
+
+This flag instructs zfs send to send those original property values rather than the current locally set values.
+
+This is useful for replicating properties from first level to second level backups, on which you'd want to locally override at least ``mountpoint`` or ``canmount`` (see the :ref:`note on property replication below <job-note-property-replication>`) but don't want to lose the original value for restore purposes.
```
problame, line 102: s/don't/do not/

docs/configuration/sendrecvoptions.rst
```diff
@@ -79,6 +79,29 @@ If ``encrypted=false``, zrepl expects that filesystems matching ``filesystems``
...
+
+Enabling this flag on the first level backup machine will cause it to send the values originally received from the backup source to the second level backup, which can perform the necessary property overrides/inherits locally for itself again. This setup can be chained infinitely.
```
problame, line 104: about this whole section: It is essentially a rough outline of how one would approach multi-tier replication with zrepl. Is it a good idea to bury it within the description of send option? Shouldn't it be in the Use Case / Tutorial section?

docs/configuration/sendrecvoptions.rst
```diff
@@ -155,5 +178,29 @@ If they are installed on the receiving side as well as the sending side, propert
...
+---------------
+
 * ``mountpoint``
 * ``canmount``
+* ``overlay``
+
+Encryption
+----------
+
+If the source datasets are encrypted but you do **not** set ``encrypted`` or ``raw`` in the :ref:`send options<job-send-options>` to ``true`` on the sending side (i.e. you do an unencrypted send) **and** you enable property replication, you will have to inherit the following properties on the :ref:`receiving side<job-recv-options>`:
```
problame, line 191: Don't use `you` here.

problame: Also, maybe explain why one would want to do this? (I suppose: receiving an unencrypted stream into an encrypted pool that has keys loaded?)

docs/configuration/sendrecvoptions.rst
```diff
@@ -155,5 +178,29 @@ If they are installed on the receiving side as well as the sending side, propert
...
+
+* ``keylocation``
+* ``keyformat``
+* ``encryption``
+
+Systemd
+-------
+
+If you use systemd, you may want to look at the properties in the `zfs-mount-generator man page <https://manpages.debian.org/buster-backports/zfsutils-linux/zfs-mount-generator.8.en.html>`_ for the ZFS releases on your machines.
```
problame, line 200: I'd prefer the openzfs man pages ( https://openzfs.github.io/openzfs-docs/man/8/zfs.8.html ) but I suppose the tool doesn't have a page there?

problame (changes requested):
Thanks for improving the docs, in particular the overview text.
I marked a few style issues.

docs/configuration/sendrecvoptions.rst
InsanePrawn: I think trying to cram every bit of knowledge like this into a use case/tutorial section will overflow it. This is advice on how to handle receiving properties in the section about enabling sending and receiving properties. Seems reasonable to me.

I would expect the tutorial sections to make heavy reference to it though.

docs/configuration/sendrecvoptions.rst
InsanePrawn: Site a->Site b
Site a is encrypted. You'll either have to send it encrypted or plain, so for distribution reasons you opt for plain.
You do want to replicate properties though for backup completeness.

This I would suggest making a tutorial section for, that goes through most of the permutations of replication sides being encrypted, `{raw,encrypted}={true,false}`, properties on/off, etc.

Tl;dr: IMHO keep this mostly about the specific zrepl feature and the underlying zfs mechanisms and reserve the tutorials section for bigger setup and use case instructions

docs/configuration/sendrecvoptions.rst
InsanePrawn: Yup, I tried.

docs/configuration/sendrecvoptions.rst
InsanePrawn: Oh, derp, rereading it, it is very tutorial-ly. My bad.
We should probably move and expand it and then reference it here, I guess.

problame: Maybe push it down to the `job-note-property-replication` section? It already contains a bunch of prose...

docs/configuration/sendrecvoptions.rst
problame: Intentional because inherit-and-override'are one level, levels being separated by `-`

docs/configuration/sendrecvoptions.rst
problame: OK, but still the `you` should go away *and* the sentence is way too long / complex / has too many conditionals.

problame:
@InsanePrawn will you address the review comments or should I do it?

problame:
Also: `com.ubuntu.zsys:bootfs`

InsanePrawn:
> @InsanePrawn will you address the review comments or should I do it?

Oh, just saw this
Not sure if I'll have time this weekend, feel free to fix them up if you find the time, your doc language seems to be better anyway; I felt I was mostly just quickly adding a little thing to help finish this PR, somewhat as a suggestion.
At this point I'll not locally stage any doc commits for this PR to refine for long; if I have something, I'll push it. (So should you for this to work I guess)

Especially when giving little doc nits as review comments on the "code", in this PR's case you could just commit or in general should do the Github inline suggestion thing IMHO :)

I'm afraid the long sentence thing would take some speech therapy to fix in general, so maybe just push/suggest some way to clean this up. 🙃 If there is excess to trim, get rid of it; it's in git anyway. I'll probably complain if something important gets lost.

problame (approved):

problame:
I'd consider this PR ready for final review / testing before we merge it.
@InsanePrawn @jwittlincohen your turn!

problame:
I have cleaned up the commit history so that it's ready for merge.
I have not tested the changes yet (still busy with #411 and testing a bump of protobuf + grpc).
@InsanePrawn any chance this will be tested until next weekend?

InsanePrawn:
Oh dear, I keep forgetting about this.
Yes, I super duper pinky swear I'll have this tested until Sunday.

InsanePrawn:
Works for me and LGTM.

jwittlincohen:
Thanks for merging this feature! Any ETA on when it may be incorporated into a stable zrepl release?

problame:
The plan is to cut 0.4-rc1 this Sunday. If you can find the time please test #297 .

PR #293, opened 2020-03-26
 endpoint: refactor, fix stale holds on initial replication, holds release subcmds

### LINKED ISSUE #278: holds: add subcmd to clean up
Due to repeated replication failures (I suspect), a single dataset built up around 400 snapshots, 95% of which were held by a single zrepl job.
Cleaning that up was not fun, as the holds (correctly) stopped me from nuking everything with a simple `destroy -r` (or at least `-d1`).

Hence, I suggest to add some form of cleanup sub command to `zrepl holds`, e.g. `zrep holds clean [$jobname]`
I'm not sure if (for every job type) multiple snapshots of the same dataset locked by a single job is a valid situation. If not, that's a situation we can detect and fix by deleting all but the newest(?) snapshot's hold of the same name/job.
Otherwise/furthermore, I'd like a `zrepl holds release [-r] $dataset [$jobname]`, which [optionally recursively] removes all holds by all or a specified zrepl job, note the mandatory dataset argument.
Useful for quickly clearing out zrepl holds before nuking something, still leaving foreign holds intact.

I strongly suspect that zrepl is in a much better position to do some smart hold cleaning for me than my brutish shell loops, even with `zrepl holds list`.

### LINKED ISSUE #280: cmd holds list: race against pruning
`zrepl holds list` shouldn't error out when a snapshot is not existent anymore, we might want to note it but keep going.

```
list holds and bookmarks on "yolopool/data/zrepltest4/yolopool/data/zrepltest3/abc/123": get holds of "yolopool/data/zrepltest4/yolopool/data/zrepltest3/abc/123@zrepl_20200219_145053_000": zfs exited with error: zfs holds failed: exit status 1
stderr:
cannot open 'yolopool/data/zrepltest4/yolopool/data/zrepltest3/abc/123@zrepl_20200219_145053_000': dataset does not exist
no datasets available
```

problame:
endpoint: refactor, fix stale holds on initial replication, holds release subcmds

- endpoint abstractions now share an interface `Abstraction`
- pkg endpoint now has a query facitilty (`ListAbstractions`) which is
  used to find on-disk
    - step holds and bookmarks
    - replication cursors (v1, v2)
    - last-received-holds
- the `zrepl holds list` command consumes endpoint.ListAbstractions
- the new `zrepl holds release-{all,stale}` commands can be used
  to remove abstractions of package endpoint

supersedes #282 
fixes #280
fixes #278


problame:
(Also, any code review is appreciated!)

problame:
- [ ] Run thison my setup for a day, see if there are any memory leaks we didn't spot.
- [ ] `ListStale` duplicates the staleness criteria for each abstraction. That should be refactored.

JMoVS:
I ran it and cleared some holds (incidentally while zrepl was running), it then complained that it couldn't replicate because the last-received-hold couldn't be moved because it has to be a snapshot.

I then cleared all holds, installed the master branch (which worked again), then tried the branch again. Even then, it currently always fails with:

`erver error: cannot move last-received-hold: last-received-hold: target must be a snapshot: seaOfTime/encRoot/SergeantPepperSenior/ocean/Users@zrepl_20200328_045822_000 `

like errors for every dataset.

The master branch works though

problame:
Which commit were you running? I pushed https://github.com/zrepl/zrepl/commit/8755847a4089fa68deaf6048d7c17365fb54bb5e to this branch yesterday evening which, despite it's wrong commit message, fixes the problem with the last-received-hold. I'd appreciate it if you could once again compare the most recent version of this branch and continue testing it.

JMoVS:
Weird, I had thought that I tried that - nevertheless I tried again and it works now. It's just not as speedy as the non-hold based version at all currently

problame:
Please let it do a few runs, then send the debug (!) logs. We are logging execution time now, I'll run some stats and see what's the bottleneck in your case.

JMoVS:
config option log level "debug" is sufficient?

problame:
Yes, just make sure the thing you are logging to doesn't throw away log messages, that would skew the stats. (I suppose it doesn't)

problame:
- [ ] only call `zfs.ZFSHolds` if we are actually requesting hold-based abstractions

problame:
- [ ] Use `userrefs` on `ZFSListFIlesystemVersions` and only visit those snapshots with > 0

problame:
- [ ] `zrepl holds release-stale` should consider the replication cursors

problame:
final revision of this PR will be reviewed here https://github.com/zrepl/zrepl/pull/300

PR #294, opened 2020-03-27
zfs: use exec.CommandContext everywhere

InsanePrawn:
Laying the groundwork to logging all /bin/zfs calls.


problame:
follow-up in #296

PR #297, opened 2020-03-27
WIP: new status terminal UI

### LINKED ISSUE #220: client/status: evaluate more abstract GUI frameworks
https://github.com/rivo/tview
https://github.com/gizak/termui

### LINKED ISSUE #245: zrepl status - Cannot scroll on long output
Hi and good afternoonm,

sorry for opening again an issue - but maybe somebody else might encounter this as well:
When you have approx <50 datasets to replicate and want to watch the progress via `zrepl watch` you cannot scroll the output and get maybe relevant information cut here:
E.g. you miss the `Pruning Sender:` or `Pruning Receiver:` blocks

Is there any workaround for except `zrepl watch --raw`? Piping or redirecting the output will not work I guess due termbox-golang is used, right? https://github.com/zrepl/zrepl/blob/master/client/status.go

problame:
Available as subcommand

    zrepl status-v2

Primary goals:

- Scrollable output ( fixes #245 )
- Sending job signals (reset, wakeup) from status view
- Filtering of output by filesystem

Implementation:

- use github.com/rivo/tview
- however, don't buy into tview too much, since not that actively
  maintained

- instead: **port over the existing status ui drawing code
  and adjust it to produce strings instead of directly
  drawing into the termbox buffer**

fixes #245
fixes #220


InsanePrawn:
Long term goal: (clearable?) error message log, e.g. for #399

problame:
I just rebased this branch onto `master`. The binaries produced by CI should be usable as a drop-in-replacement for zrepl 0.3.* with no major changes compared to 0.3.1. Please play around with `zrepl status-v2` and report feedback.

problame:
- [ ] visualize `.Global`, in particular the currently executing ZFS commands. The information is already in the JSON, see `zrepl status --raw`

calistoc:
just checked a little the new “zrepl status-v2” building the status-v2 branch, it looks good to me and it is nice we can now scroll when there are many lines, thank you.

I tried to signal a job with Shift+S and when you get to decide whether to wakeup or reset a job there is no way to go back, also in case you use ESC it exits with a panic

panic: runtime error: index out of range [-1]
goroutine 1517 [running]:
github.com/zrepl/zrepl/client/status%2ev2.Main.func12.1.1(0xc0000a8240, 0x2, 0x2, 0xffffffffffffffff, 0xc0004aef30, 0xc00046e800, 0xc00025f020)
/src/client/status.v2/main.go:274 +0x10e
created by github.com/zrepl/zrepl/client/status%2ev2.Main.func12.1
/src/client/status.v2/main.go:273 +0x75

problame:
Will look into the crash later today.

- have you tried the search filter functionality? Is it sufficiently intuitive?
- should we ship this as a replacement for the current status subcommand? I don't see any downsides apart from the sidebar requiring a slightly wider terminal

calistoc:
> Will look into the crash later today.

thanks

> 
>     * have you tried the search filter functionality? Is it sufficiently intuitive?

yes it looks ok and useful to me

> 
>     * should we ship this as a replacement for the current status subcommand? I don't see any downsides apart from the sidebar requiring a slightly wider terminal

yes I think the status-v2  should replace the current status subcommand though I would leave it there somehow, like adding a --dump flag to show all the jobs to the terminal, so no interactive, just an idea.

my two cents here, if possible regarding the status-v2 subcommand:

- I would give the possibility to switch panes using the arrows and maybe removing the TAB completely
- For each job I would also add the Job name just above the Type, like:
Job: server1_to_server2
Type: push
...
this should also help with the sidebar size as you might always end up with longer job names that won't fit there

- On the main job pane I would add a "last run timestamp" for each section, like:
Job: server1_to_server2
Type: push
Replication:
Last run: 2021-...
..
Pruning Sender:
Last run: 2021-...
..
Pruning Receiver:
Last run: 2021-...
..

thank you

problame:
- [x] add the `status-v2 --mode {raw,dump,interactive}` flag
- [x] fix the panic  https://github.com/zrepl/zrepl/pull/297#issuecomment-753012773
- [x] add the job name

> 
> On the main job pane I would add a "last run timestamp" for each section, like:
> Job: server1_to_server2
> Type: push
> Replication:
> Last run: 2021-...
> ..
> Pruning Sender:
> Last run: 2021-...
> ..
> Pruning Receiver:
> Last run: 2021-...

- [ ] Hm, we have similar needs for the prometheus metrics ( cc #353 ). We should probably add this as a field that is computed server-side so all consumers of the status (struct or JSON) can use it.

> I would give the possibility to switch panes using the arrows and maybe removing the TAB completely

I think the original idea was to reserve left-right arrows for folding, but we won't have folding without a lot more buy-in into the terminal ui library, which I'm not ready to do.
Either way, I would find it weird if inter- and intra-pane navigation both happened with the same shortcuts.

problame:
User feedback:

- [ ] Would like some kind of overview screen for all jobs (e.g. status, last run, etc)
  - probably also useful as a stand-alone command, e.g. `zrepl status summary`
- [ ] `ESC` in filter input mode doesn't leave the filter 
- [ ] filter seems to filter progress as well => non-matching filter causes negative replication speed
- [ ] specifying `--job` should imply `--mode dump`. Alternatively, would expect it to pre-select the right job in `--mode interactive`. 
- [ ] **strong** `q` doesn't quit the tool (unlike top or htop), would have expected that
- [ ] tab only works in one direction (no shift tab?)

calistoc:
>     * [ ]  Would like some kind of overview screen for all jobs (e.g. status, last run, etc)
>       
>       * probably also useful as a stand-alone command, e.g. `zrepl status summary`
> 

regarding the Replication status I think this could be a quick way to add the Last Run information

```diff
--- a/client/status.v2/viewmodel/render.go
+++ b/client/status.v2/viewmodel/render.go
@@ -321,6 +321,10 @@ func renderReplicationReport(t *stringbuilder.B, rep *report.Report, history *by
 
        t.Printf("Status: %s", latest.State)
        t.Newline()
+       if !latest.FinishAt.IsZero() {
+               t.Printf("Last Run: %s (lasted %s)\n", latest.FinishAt, latest.FinishAt.Sub(latest.StartAt))
+       }
```

I tried this and it seems to be working fine

calistoc:
one additional feedback,
not sure if you already noticed this but I see the zrepl status-v2 eating up cpu cycles while monitoring the replications, up to 15%-20% cpu usage especially when the replication is actively copying data, for what I saw that is caused by the for range loop in the status_interactive.go with t := time.NewTicker(300 * time.Millisecond)

Increasing the ticker to 1000ms helped a lot with the cpu usage that went down, having 1 sec refresh also gives a better idea of the replication progress speed as it gets more stable and easy to read.
Just a suggestion, I hope it helps.
Thank you

problame:
> one additional feedback,
> not sure if you already noticed this but I see the zrepl status-v2 eating up cpu cycles while monitoring the replications, up to 15%-20% cpu usage especially when the replication is actively copying data, for what I saw that is caused by the for range loop in the status_interactive.go with t := time.NewTicker(300 * time.Millisecond)
> 
> Increasing the ticker to 1000ms helped a lot with the cpu usage that went down, having 1 sec refresh also gives a better idea of the replication progress speed as it gets more stable and easy to read.
> Just a suggestion, I hope it helps.
> Thank you

Yeah, the 300ms might be a little too aggressive ( @lapo-luchini [is having trouble with it as well with the old status command, and it seems to impact the daemon's replication performance as well](https://github.com/lapo-luchini) )

- [ ] Integrate @calistoc 's patch posted above
- [ ] increase the default ticker interval to some higher value (e.g. 1  second)
- [ ] make ticker interval configurable via command line flag

@calistoc Can I interest you in finishing up this PR? I'd like to ship it as a preview in zrepl 0.4 and see how people react.

calistoc:
Happy to help if I can, I have already updated my status-v2 branch with the above changes plus some minor ones.
On the same branch though I also added in the status-v2 UI the option to trigger the job snapshotting:

via status-v2 UI using Shift-S, then select the “snapshot” option.
Or
$ zrepl signal snapshot JOB

I have been testing this for a while and all seems ok with the status-v2, it is also very useful to have the possibility to snap directly via zrepl UI, for example before any major update that right now would require manual snapshots outside zrepl. Having the option to snap via zrepl I managed to get rid of other snapshotting scripts.

What I’m not sure though is if you are interested about the UI/cli snapshot trigger as well, I do hope so anyway.

I can work on this and propose the code to you, should I open a new PR on the status-v2 branch?

Please bear in mind that I’m not that expert with Go and just started using it about two months ago, so double check my code :)

problame:
> Happy to help if I can, I have already updated my status-v2 branch with the above changes plus some minor ones.
> On the same branch though I also added in the status-v2 UI the option to trigger the job snapshotting:
> 
> via status-v2 UI using Shift-S, then select the “snapshot” option.
> Or
> $ zrepl signal snapshot JOB

Cool! Did you change the daemon-side to implement this? Currently the wakeup signal only wakes up the replication bits of a push job.

> I can work on this and propose the code to you, should I open a new PR on the status-v2 branch?
> Please bear in mind that I’m not that expert with Go and just started using it about two months ago, so double check my code :)

Yes, I won't touch the status-v2 branch so you can use it as a merge-base for your PR.
No worries about being new to Go and have fun!

calistoc:
yes I had to change the daemon-side a little, the signal snapshot does also trigger the replication in case of push jobs
just opened new PR #418

problame:
Great! I just did a quick review. Let's continue the discussion there.

problame:
- [x] merged all of @calistoc 's changes from #418 
- [x] hotkey `q` to quit the interactive UI
- [x] immediate redraw on wrap
- [ ] attempt upgrade of tview to newer version (current version is more than a year old)
  - reverted that commit because of tview pr https://github.com/rivo/tview/pull/523 (border background color is broken)
  - tview also changed the background colors. Not sure to what they changed them exactly, maybe this is also fallout from the tcell update (transient dependency of tview).
  - See [here](https://github.com/rivo/tview/pull/523#issuecomment-778837580) and [here](https://github.com/rivo/tview/pull/523#issuecomment-778838077) for comparison.
  - [ ] => decide on how we want to deal with background color in general
    - I'd like a `light` mode and a `dark` mode at some point in the future but I'd be fine merging this PR with _one_.
      - `zrepl status` didn't have any color and just used foreground / background colors => always fit perfectly into the user's terminal emulator color profile
      - can we have the same for `status-v2`?

InsanePrawn:
>     * [x]  merged all of @calistoc 's changes from #418

🥳 
>     * [ ]  attempt upgrade of tview to newer version (current version is more than a year old)
>       
>       * reverted that commit because of tview pr [rivo/tview#523](https://github.com/rivo/tview/pull/523) (border background color is broken)
>       * tview also changed the background colors. Not sure to what they changed them exactly, maybe this is also fallout from the tcell update (transient dependency of tview).
>         * I'd like a `light` mode and a `dark` mode at some point in the future but I'd be fine merging this PR with _one_.

my thoughts in your words:

>           * `zrepl status` didn't have any color and just used foreground / background colors => always fit perfectly into the user's terminal emulator color profile
>           * can we have the same for `status-v2`?

all of the ncurses(-like) tools i can think of right now seem to follow this convention (by default at least where theming is allowed, especially editors): {,h,io}top, pulsemixer (edit: alsamixer does set a bg :D), weechat, ncdu, pagers, editors, ...

If someone really has a strong need for this, we could implement three schemes: native, dark, light, where only the latter two override the default background and foreground colours for 'uncoloured' text/areas.

A-ha! I believe I found the robots we're looking for: https://github.com/termoose/irccloud/blob/01c817df66bed38bc7c0a8000001579efccd4c71/main.go#L16

Quick PoC: InsanePrawn/zrepl@a53ea77

InsanePrawn:
Note: also works for me without reverting the update.
InsanePrawn/zrepl@a1f60ac

InsanePrawn:
Meh, I'm confused.
InsanePrawn/zrepl@4eb6e7b and InsanePrawn/zrepl@49c2d8f (with and without the revert) both seem to work correctly for me now on both terminal color schemes.

Not sure if this is known: For me, switching jobs only partially resets/redraws the job view pane. Areas not touched by the new job (i.e. its string representation is shorter than the last displayed job's) will retain the old text content.
Both branches experience this bug.

problame:
> Edit2: It seems your git acrobatics didn't go so well, the signal wakeup changes are in here too. :) 
yes they are but they have been moved to a commit prior to the introduction of status.v2

problame:
> Not sure if this is known: For me, switching jobs only partially resets/redraws the job view pane. Areas not touched by the new job (i.e. its string representation is shorter than the last displayed job's) will retain the old text content.
Both branches experience this bug.
Edit: Seems it only needs a redraw, resizing the terminal fixes the display.

@InsanePrawn tview PR https://github.com/rivo/tview/pull/395 fixes this.

problame:
- rebased this branch onto latest master. This now includes the send-recv-flags work ( https://github.com/zrepl/zrepl/pull/285 , off by default ) and several concurrency related fixes.
- pulled in @InsanePrawn 's DefaultColor patches
- experimented a bit with https://github.com/rivo/tview/pull/395
- reverted that experiment
- experimented with https://gitlab.com/tslocum/cview (a fork of tview)
  - fixes the DefaultColor + SetText bug
  - we now use tcell.ColorGreen to highlight selections, I'd prefer 'bold' vs 'not-bold' but that seems to be hard -,-
  - regardless, I think this is in a presentable shape

Please continue to test this branch and provide feedback.

problame:
- [ ] error alerts still have the wrong color
  image

InsanePrawn:
the windows nostalgia :scream:

InsanePrawn:
Uhm, maybe I should've shared this before: https://github.com/rivo/tview/blob/master/styles.go
InsanePrawn/zrepl@a61b75e

problame:
I implemented a `status-v2 --mode v1` which imitates `status`. Do you think we can get away with dropping the status-v1 code and shipping `status` as an alias to `status-v2 --mode v1`? (i.e. will users notice the difference?)

docs/changelog.rst
```diff
@@ -27,6 +27,16 @@ We use the following annotations for classifying changes:
...
 
+0.4.0
+-----
+
+* |break| Change syntax to trigger a job replication, rename ``zrepl signal wakeup JOB`` to ``zrepl signal replication JOB``
+* |feature| New ``zrepl status`` UI:
+
+  * Interactive job selection.
+  * Interactively ``zrepl signal`` jobs.
+  * Filter filesystems in the job view by name.
```
InsanePrawn, line 38:
```suggestion
  * Filter filesystems in the job view by name.
  * An approximation of the old UI is still included as `--mode legacy`
```

InsanePrawn (approved):

problame:
Merged! Thanks @calistoc for your work on this and to everyone who helped test this feature!

PR #296, opened 2020-03-27
zfs command logging, status, prometheus metrics

problame:

PR #300, opened 2020-04-07
endpoint: refactor, fix stale holds on initial replication failure, holds release subcmds, more efficient ZFS queries

### LINKED ISSUE #278: holds: add subcmd to clean up

### LINKED ISSUE #280: cmd holds list: race against pruning

problame:
…olds release subcmds, more efficient ZFS queries

The motivation for this recatoring are based on two independent issues:

- @JMoVS found that the changes merged as part of #259 slowed his OS X
  based installation down significantly.
  Analysis of the zfs command logging introduced in #296 showed that
  `zfs holds` took most of the execution time, and they pointed out
  that not all of those `zfs holds` invocations were actually necessary.
  I.e.: zrepl was inefficient about retrieving information from ZFS.

- @InsanePrawn found that failures on initial replication would lead
  to step holds accumulating on the sending side, i.e. they would never
  be cleaned up in the HintMostRecentCommonAncestor RPC handler.
  That was because we only sent that RPC if there was a most recent
  common ancestor detected during replication planning.
  @InsanePrawn prototyped an implementation of a `zrepl holds release`
  command to mitigate the situation.
  As part of that development work and back-and-forth with @problame,
  it became evident that the abstractions that #259 built on top of
  zfs in package endpoint (step holds, replication cursor,
  last-received-hold), were not well-represented for re-use in the
  `zrepl holds release` subocommand.

This commit refactors package endpoint to address both of these issues:

- endpoint abstractions now share an interface `Abstraction` that, among
  other things, provides a uniform `Destroy()` method.
  However, that method should not be destroyed directly but instead
  the package-level `BatchDestroy` function should be used in order
  to allow for a migration to zfs channel programs in the future.

- endpoint now has a query facitilty (`ListAbstractions`) which is
  used to find on-disk
    - step holds and bookmarks
    - replication cursors (v1, v2)
    - last-received-holds
  By describing the query in a struct, we can centralized the retrieval
  of information via the ZFS CLI and only have to be clever once.
  We are "clever" in the following ways:
  - When asking for hold-based abstractions, we only run `zfs holds` on
    snapshot that have `userrefs` > 0
    - To support this functionality, add field `UserRefs` to zfs.FilesystemVersion
      and retrieve it anywhere we retrieve zfs.FilesystemVersion from ZFS.
  - When asking only for bookmark-based abstractions, we only run
    `zfs list -t bookmark`, not with snapshots.
  - Currently unused (except for CLI) per-filesystem concurrent lookup
  - Option to only include abstractions with CreateTXG in a specified range

- refactor `endpoint`'s various ZFS info  retrieval methods to use
  `ListAbstractions`

- change the `zrepl holds list` command to consume endpoint.ListAbstractions

- Add a `ListStale` method which, given a query template,
  lists stale holds and bookmarks.
  - it uses replication cursor has different modes
- the new `zrepl holds release-{all,stale}` commands can be used
  to remove abstractions of package endpoint

- Adjust HintMostRecentCommonAncestor RPC for stale-holds cleanup:
    - send it also if no most recent common ancestor exists between sender and receiver
    - have the sender clean up its abstractions when it receives the RPC
      with no most recent common ancestor, using `ListStale`
    - Due to changed semantics, bump the protocol version.

- Adjust HintMostRecentCommonAncestor RPC for performance problems
  encountered by @JMoVS
    - by default, per (job,fs)-combination, only consider cleaning
      step holds in the createtxg range
      `[last replication cursor,conservatively-estimated-receive-side-version)`
    - this behavior ensures resumability at cost proportional to the
      time that replication was donw
    - however, as explained in a comment, we might leak holds if
      the zrepl daemon stops running
    - that  trade-off is acceptable because in the presumably rare
      this might happen the user has two tools at their hand:
    - Tool 1: run `zrepl holds release-stale`
    - Tool 2: use env var `ZREPL_ENDPOINT_SENDER_HINT_MOST_RECENT_STEP_HOLD_CLEANUP_MODE`
      to adjust the lower bound of the createtxg range (search for it in the code).
      The env var can also be used to disable hold-cleanup on the
      send-side entirely.

supersedes closes #293
supersedes closes #282
fixes #280
fixes #278

Additionaly, we fixed a couple of bugs:

- zfs: fix half-nil error reporting of dataset-does-not-exist for ZFSListChan and ZFSBookmark

- endpoint: Sender's `HintMostRecentCommonAncestor` handler would not
  check whether access to the specified filesystem was allowed.


problame:
- [ ] Run this 24h on @problame 's setup
- [ ] @JMoVS feedback on performance
  - [ ] please experiment with the mentioned env variable (name changed compared to prior revisions of this PR, read the code for possible values!)
- [ ] @InsanePrawn confirm bugfix for leaking step holds

JMoVS:
I'll compile and check - might take me a week though

problame:
- [ ] Should `zrepl holds release-all` be dry-run be default?

JMoVS:
>     * [ ]  Should `zrepl holds release-all` be dry-run be default?

I think if we followed zfs "normal", we could say that zfs destroy doesn't need a "yes I really want" switch. But dry run does. Only when an unintended side effect (eg overwriting snapshots, so "force" is needed should we consider this.

So for me the question is: Isn't the only thing potentially lost a resumable send state because aside from the holds, there are still the replication cursors making incremental send possible at all times?

problame:
> So for me the question is: Isn't the only thing potentially lost a resumable send state because aside from the holds, there are still the replication cursors making incremental send possible at all times?

- [ ] Well, `zrepl holds` should actually be called `abstractions`, as it encompasses _all_ abstractions created by zrepl. Have a look at the `zrepl holds release-all --help --types` flag. Or maybe we should have separate subcommands for each abstraction type, i.e. `zrepl abstraction release-{stale,all} {step,replication-cursor,last-received-hold`
- [] find a catchier name for the abstractions than `zrepl abstraction`

JMoVS:
I have now tested all 3 options for the environment variable and speeds are only marginally different. When looking at zrepl status, it spends the most time at 0B/s transfer rate. For the new version to become speedy again, I think we would either need to revert/comment out the holds logic or we'd need to also put in parallel replication.

problame:
Parallel replication is available via an env var in #303 which includes the changes in this PR. Could you please experiment with that?

problame:
Follow-up (sry, was on mobile): the commit in #303 that adds env-var-configurable parallel replication is 632acb6305d0ff1436b3e3938c2fa99e0fd8a707 - you should be able to cherry-pick it ontop of this PR, or run #303 instead of this PR.

problame:
Summary of @JMoVS  's experiments with this branch:

* OS X
* `ZREPL_ENDPOINT_SENDER_HINT_MOST_RECENT_STEP_HOLD_CLEANUP_MODE=range-since-replication-cursor` (the default)
* `zfs holds` is no longer the bottleneck, the majority of time zrepl spents waiting for zfs commands is `zfs get` on the sending-side pool's snapshots.
* a seaborn boxplot of the `zfs get` latencies per dataset shows similar results for each dataset:
  * Q1 at `~0.075s`
  * Q3 at `~0.150s`
  * Significant amount of outliers in the `[0.300s,3.000s]` interval
  * Fewer but still significant amount of outliers in the `[3.000s,10.000s]` interval
  * 1 30s outlier
  * 1 55s outlier
  * 1 150s outlier
* The sending-side pool is being actively used
* There is another tool that does periodic `zfs list -t snapshot -r -d 1` for every filesystem on the sending-side pool.
* Possible future analysis steps:
  * determining whether the `dmu_hold_objset` or `zfs_ioc_objset_stats_impl`
    * see bpftrace and dtrace scripts for this here https://gist.github.com/problame/4f24d5506ad280f948dfa250e22b1a17
    * those still need to be ported to OS X, @problame  doesn't have an OS X machine, so that's @JMoVS  's job to figure out
  * once #307 is done, use Chrome Trace Event Profile to analyze time spent during replication

problame:
Decision:

* Merge this branch after the naming issue for the `holds` subcommand is resolved, despite @JMoVS 's performance problems.
* Don't make a release 0.3 until #307 is done, which provides a neat visualization of where zrepl spends its time:
  image
* Use #307 to pin down @JMoVS 's perf problems.

problame:
Decision on the name of the sub-command: `zrepl zfs-abstraction {list,release-stale,release-all}`.
Rationale: these commands should not need to be used regularly in the first place, and shell autocompletion will do the rest.

Issue #219, opened 2019-09-21
docs: section for typical / desired use cases (tutorial style)


problame:
- [ ] Encourage users to write such tutorials and submit them as PRs

### Topics

- [ ] Using Hooks for consistent MySQL database backups
- [ ] Backing up ZoL 0.8 with native encryption to $DESTINATION_SUPPORTING_RAW_RECV
  - show off how nice encryption support is #218 
  - maybe ZoL people are more likey to use it on their laptop? => show off resumable send & recv #217 
  - preferrably a "popular" destination, e.g. FreeNAS
- [ ] Backing up ZFS on FreeBSD (ZoF) with 
  - yes, this is basically SEO to get new users looking for encryption support :P
- [ ] Backing up your systems (Mac!) to FreeNAS using {push, pull} replication

problame:
Call this done after we point out options for encryption, etc, in the docs

PR #312, opened 2020-04-21
Add FreeBSD jail tutorial

johnramsden:
# Summary

Add a tutorial page https://github.com/zrepl/zrepl/issues/219 instructing how to set up a FreeBSD jail using `iocage`.

## Details

I have left instructions for both using a iocage plugin and for a manual install. Let me know if you think it makes more sense to remove the plugin instructions

For the plugin, it has been added to the [ix-plugin-hub](https://github.com/ix-plugin-hub/iocage-plugin-index/blob/master/zrepl.json), note the plug-in has not yet been merged into a release branch https://github.com/ix-plugin-hub/iocage-plugin-index/pull/71#issuecomment-617449084, at which point it will be available via the FreeNAS UI and cli. Until then I have left instructions on how to use the plugin by downloading the json file directly. Once it has been added, following command can be used instead of `L106-107`:

    iocage fetch -P zrepl.json --name zrepl jail_zfs_dataset=zrepl boot=on

The plug-in basically runs the steps that are manually run by the user, setting up logging, enabling the service to start on boot, and showing a post install message - see [post_install.sh](https://github.com/johnramsden/iocage-plugin-zrepl/blob/master/post_install.sh).

Let me know if you would be interested in me transferring the artifact repo [johnramsden/iocage-plugin-zrepl](https://github.com/johnramsden/iocage-plugin-zrepl) over to @zrepl, it doesn't matter to me either way.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,138 +1,19 @@
-.. _binary releases: https://github.com/zrepl/zrepl/releases
 
-.. _installation:
+.. _installation_toc:
 
+************
 Installation
-============
+************
 
 .. TIP::
 
     Note: check out the :ref:`tutorial` if you want a first impression of zrepl.
 
-User Privileges
----------------
+.. toctree::
 
-It is possible to run zrepl as an unprivileged user in combination with
-`ZFS delegation <https://www.freebsd.org/doc/handbook/zfs-zfs-allow.html>`_.
-Also, there is the possibility to run it in a jail on FreeBSD by delegating a dataset to the jail.
-However, until we get around documenting those setups, you will have to run zrepl as root or experiment yourself :)
-
-Packages
---------
-
-zrepl source releases are signed & tagged by the author in the git repository.
-Your OS vendor may provide binary packages of zrepl through the package manager.
-Additionally, `binary releases`_ are provided on GitHub.
-The following list may be incomplete, feel free to submit a PR with an update:
-
-.. list-table::
-    :header-rows: 1
-
-    * - OS / Distro
-      - Install Command
-      - Link
-    * - FreeBSD
-      - ``pkg install zrepl``
-      - `<https://www.freshports.org/sysutils/zrepl/>`_
-    * - MacOS
-      - ``brew install zrepl``
-      - Available on `homebrew <https://brew.sh>`_
-    * - Arch Linux
-      - ``yay install zrepl``
-      - Available on `AUR <https://aur.archlinux.org/packages/zrepl>`_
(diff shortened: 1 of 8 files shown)


johnramsden:
@problame Sounds good, making the adjustments.

Regarding adding as a child node of installation. Should I do as you have done with configuration and add a installation directory, and installation table of contents page. I [don't see](https://www.sphinx-doc.org/en/latest/usage/restructuredtext/directives.html#table-of-contents) a way to add the page _only_ to the sidebar.

problame:
Yes, I think that's a limitation of sphinx

johnramsden:
Additionally, when adding it to the installation table, should I just add a second line for FreeBSD with the link? That works but it makes it no longer even with the size of the other rows.

johnramsden:
- [x] I think it should be pkg install py37-iocage
- [x] iocage setup steps not fully described
- [x] also, hardcoding the FreeBSD release seems odd to me. can we use `$(freebsd-version -u)` or `$(freebsd-version -k)` instead?

  It has to be without the patch set, so I used `$(freebsd-version -k | cut -d '-' -f '1,2')`

- [x] the 'plugin' and 'manual' jail creation section should probably be in the same order as the outline section that precedes them
- [ ] the 'create log file' step in the manual section should be covered by the pkg installer, right? (cc @woodsb02 )

  Upon install, the package instructions say to manually run through these steps, perhaps it's not necessary?
- [x] relocate this tutorial to a child node of the docs' 'Installation' section (titled "FreeBSD Jail With `iocage`")

  I went ahead and changed the hierarchy adding individual pages for installation, otherwise the "What Next" section appeared before the "FreeBSD Jail With iocage" page which seemed a little odd. Let me know if there's another way you would like this organized.
- [x] refer to it from the packages table

  As mentioned above, the table looks a little oddly shaped now, let me know what you think.

problame:
> I went ahead and changed the hierarchy adding individual pages for installation, otherwise the "What Next" section appeared before the "FreeBSD Jail With iocage" page which seemed a little odd. Let me know if there's another way you would like this organized.

:+1:

problame:
Thanks for your work on this!

johnramsden:
@problame I was just wondering when you thought this might be active on the website?

problame:
It will be available with the 0.3 release.
Docs build from master, and I usually only update the docs on every release (releases are tags on master ATM).

0.3, or at least a release-candidate for it, should be ready within one or two week's time, that's when I'll update the docs.

PR #336, opened 2020-06-20
Fix example pruning config

### LINKED ISSUE #335: Documentation: missing config in prune example?
At https://zrepl.github.io/v0.3.0-rc1/configuration/prune.html it says:

> **A snapshot that is not kept by any rule is destroyed**

Then it gives the following example:

```
    pruning:
      keep_sender:
...
        # make sure manually created snapshots by the administrator are kept
        - type: regex
          regex: "^manual_.*"
...
      keep_receiver:
        - type: grid
          grid: 1x1h(keep=all) | 24x1h | 35x1d | 6x30d
          regex: "^zrepl_.*"
        # manually created snapshots will be kept forever on receiver
```

However, looking at the very last line, I cannot see why keep_receiver will keep manually-created snapshots.  Are there a couple of lines missing? i.e. should it say

```
        # manually created snapshots will be kept forever on receiver
        - type: regex
          regex: "^manual_.*"
```

This would be consistent with the subsequent warning:

> Make sure to actually add the necessary regex keep rules on both sides, like with manual in the example above.

candlerb:
Fixes #335

### DIFF
--- a/docs/configuration/prune.rst
+++ b/docs/configuration/prune.rst
@@ -46,6 +46,8 @@ Example Configuration:
              grid: 1x1h(keep=all) | 24x1h | 35x1d | 6x30d
              regex: "^zrepl_.*"
            # manually created snapshots will be kept forever on receiver
+           - type: regex
+             regex: "^manual_.*"
 
 .. DANGER::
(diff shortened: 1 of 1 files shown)


problame:
Thanks for this PR, going to merge the commit as part of the docs-rollup for the 0.3 release!

Issue #338, opened 2020-06-20
Prometheus metrics: gauge for current failed syncs


candlerb:
zrepl exposes counters for logs at different levels:

```
zrepl_daemon_log_entries{level="error",zrepl_job="backup_storage1"} 9
```

Whilst this is reasonably useful, I think it would be an improvement to have *gauges* for each job showing the number of filesystems which are currently out of sync, i.e. where the most recent replication attempt was a failure.

A dashboard could then show with confidence "this job is / is not in sync", rather than "an error occurred 3 hours ago, but we don't know whether it's still a problem or has been resolved"

I couldn't find an existing metric which shows this.

problame:
Thanks for this suggestion, I like the idea very much and would welcome pull requests.
If anyone is working on this, please let me know in a comment!

---

### Impl Advice

* package replicationg generates a report
* the report is what is used to render the `zrepl status` ui
* Use the `zrepl status --raw` to get an idea of how the data looks like
* Option A (preferred): implement a Getter on `github.com/zrepl/zrepl/replication/report.Report` ( https://github.com/zrepl/zrepl/blob/master/replication/report/replication_report.go#L8 ) that returns the gauge value. I can do the rest of the plumbing to expose the prometheus metrics.
* Option B: Look at the  `zrepl status --raw` JSON output and provide code in `jq` / `python` / whatever that computes the gauge value out of it. I'll translate it to Go code then. (More time consuming for me, less likely to happen / more overhead)

problame:
implemented in #341 , available in zrepl 0.3 (released)

PR #341, opened 2020-06-28
Implement prometheus gauge for failed filesystem replications

schulzh:
Implements the proposed prometheus gauge in https://github.com/zrepl/zrepl/issues/338
I'm not entirely sure what to return in `GetFailedFilesystemsCount` in these cases:
- Replication is not done yet -> AFAICS it does not matter, since the prometheus gauge is only set after replication exits, so I opted for `0`
- Replication fails and filesystem reports are still empty -> I opted for `-1` and an explaination in the metric description

Let me know what you think.

I have tested this locally::
- target not reachable -> `-1` as expected
- receiver interrupted while sending -> correct count
- successful replication -> 0
- multiple jobs -> works correctly

### DIFF
--- a/replication/report/replication_report.go
+++ b/replication/report/replication_report.go
@@ -162,3 +162,30 @@ func (f *FilesystemReport) NextStep() *StepReport {
 func (f *StepReport) IsIncremental() bool {
 	return f.Info.From != ""
 }
+
+// Returns, for the latest replication attempt,
+// 0  if there have not been any replication attempts,
+// -1 if the replication failed while enumerating file systems
+// N  if N filesystems could not not be replicated successfully
+func (r *Report) GetFailedFilesystemsCountInLatestAttempt() int {
+
+	if len(r.Attempts) == 0 {
+		return 0
+	}
+
+	a := r.Attempts[len(r.Attempts)-1]
+	switch a.State {
+	case AttemptPlanningError:
+		return -1
+	case AttemptFanOutError:
+		var count int
+		for _, f := range a.Filesystems {
+			if f.Error() != nil {
+				count++
+			}
+		}
+		return count
+	default:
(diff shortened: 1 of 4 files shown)


replication/report/replication_report.go
```diff
@@ -162,3 +162,30 @@ func (f *FilesystemReport) NextStep() *StepReport {
 func (f *StepReport) IsIncremental() bool {
 	return f.Info.From != ""
 }
+
+// Returns, for the latest replication attempt,
+// 0  if there have not been any replication attempts,
```
schulzh, line 167: Is this a valid situation, a replication report with no attempts? If yes, I'd consider it to be wrong to not report an error, since nothing was replicated, or am I missing something here?
Also, 0 is returned if everything was successfully replicated.

problame: > Is this a valid situation, a replication report with no attempts? If yes, I'd consider it wrong to not report an error, since nothing was replicated, or am I missing something here?

When the daemon starts up, a push job will have 0 replication attempts reported until it is woken up via command line or because its snapshotter created a snapshot.

problame:
@schulzh Have you used the current state of this branch on your systems?

schulzh:
Sorry for the late reply. I tested it again (after updating the branch) and it works as expected.
I can also provide a dynamic Grafana panel showing the current state of all configured replication jobs if wanted as a second pr:
image

problame:
Great to see progress on this nonetheless!
Could you just add your changes to `dist/grafana/grafana-prometheus-zrepl.json` as part of this PR?

problame:
Thanks again for following up on this!

I added a changelog entry and squashed everything together, will merge to master once CI is done.
This change is going to be part of 0.3

PR #351, opened 2020-08-09
zrepl Dockerfile

enspritz:
G'Day

Investing some effort into creating a Docker image capable of running the `zrepl` daemon in a Docker container.
Hitting roadblocks and would appreciate some input/wisdom, but first:
Are you receptive to this contribution? :D


enspritz:
In its current form, the `Dockerfile` builds an image which can be run.
If not provided with our own configuration file, the `zrepl` daemon fails with reason ("no jobs" configured), the container exits, and we can read why using `docker logs`. This is expected, and instructive for seasoned Docker operators new to the zrepl image:
```bash
$ sudo docker run -d --name zrepl zrepl:0.2.1
$ sudo docker logs -f zrepl
could not parse config: yaml: unmarshal errors:
  line 0: field jobs is required but has zero value
```


But if provided with a good configuration, `zrepl` is appears silent to me, and more alarmingly doesn't write its socket file, making `zrepl status` unusable:
```bash
$ sudo docker run -d --name zrepl -v $(pwd)/my-zrepl-conf:/etc/zrepl:ro zrepl:0.2.1
$ sudo docker exec -it zrepl zrepl status
Post http://unix/status: dial unix /var/run/zrepl/control: connect: no such file or directory
```

enspritz:
I feel that the exercise of Docker-izing zrepl will long-term exert a positive influence on the behavior of the daemon, such as with regards to configuration possibilities and logging.

problame:
Hi, sorry for my late reply!

> I feel that the exercise of Docker-izing zrepl will long-term exert a positive influence on the behavior of the daemon, such as with regards to configuration possibilities and logging.

Agreed.

> 
> Investing some effort into creating a Docker image capable of running the zrepl daemon in a Docker container.
> Hitting roadblocks and would appreciate some input/wisdom, but first:
> Are you receptive to this contribution? :D

My main question is: why would you dockerize a daemon like zrepl?

* it needs access to `/dev/zfs` (and probably needs to run as `--privileged`?)
* it is a single Go binary
* it ships with a systemd service definition

I suppose you prefer docker over systemd as a service + log manager?

enspritz:
> Hi, sorry for my late reply!

Not late, you're perfectly on time ^_^

> why would you dockerize a daemon like zrepl?

> I suppose you prefer docker over systemd as a service + log manager?

Yes. I also have perfectly selfish reasons that motivate me, in case you were worried: The management, metrics, and security systems I have on hand here are tuned to Docker containers, including for monitoring and log collection, making the Docker route the Obvious Choice and the Path of Least Resistance. Dockerization is a win for a certain class of curious person like myself, as trying zrepl would be a matter of copying & pasting a single line from the README `docker run ...` and seeing zrepl get right to work, representing a lower barrier to adoption that would have saved me precious time.

I am grateful for the APT repository and the systemd service; this represents sound and much-appreciated sysadmin thinking.

enspritz:
The Docker image will have a default behavior; let us decide upon a reasonable first approximation of what that behavior aught to be.

**Configuration**: The Wielder of the Docker Image ultimately needs to supply their own zrepl.yml configuration. In Docker's reckoning, this would be a matter of attaching a Docker volume that contains the configuration file `--volume /my/zrepl/conf/dir:/etc/zrepl:ro`. Of course, one can always build `FROM` the standard zrepl Docker image to produce their own customized zrepl Docker image, `COPY`ing in their own `/etc/zrepl/zrepl.yml` file or whatever.

**Logging**: The default `zrepl.yml` included in the default Docker image should be configured for logging to `stdout` and/or `stderr`, where Docker expects to collect the logging output.

**Snapshots**: Here I have question marks for the default config: We could go with snapshots of `<` on a schedule and similar retention period of `1h` or so which shows off zrepl, bringing first impressions to at least parity with the likes of znapzend and friends. Encouragement to proceed with investigation of zrepl, if you will. These snapshots, perhaps uninvited, are easily cleaned, either manually or by zrepl itself upon configuration with different options. A quickly mention in the docs would help. Alternatively, the default config would not perform snapshots at all.

**Pull/push**: This is entirely defined by the operator, so perhaps there is nothing that can be effectively demonstrated here with universal appeal & appreciation?

Summarizing, either zrepl shows off its snapshotting abilities, or, with no `jobs` defined, the container run fails with a logging message that prods the user to provide a valid configuration file. What is a good first user experience?

In any case, I feel it would be helpful and instructive to include minimal configurations for snapshotting, server, push, pull, just the core functionality. All but logging and perhaps snapshotting would be commented-out and thus inactive. These sections assist the user to rapidly adapt zrepl to their use case. Less guesswork, and provides a solid footing into spelunking the documentation.

problame:
Thanks for those detailed strategic ideas.

> Summarizing, either zrepl shows off its snapshotting abilities, or, with no jobs defined, the container run fails with a logging message that prods the user to provide a valid configuration file. What is a good first user experience?

That's a question I have been thinking about for a while.
I don't like the idea that a zrepl Docker container would just start running and do some default snapshot management on every dataset it sees. Mostly because it might disturbing existing snapshot management tooling / intervals / etc.

And we  obviously can't have any replication functionality enabled by default:
* push: we don't know where to push and don't know what filesystems we'd be allowed to push
* pull: don't know where to connect to, don't know the `root_fs`
* source: don't know which clients should be allowed, don't know what filesystems to serve
* sink: don't know the `root_fs`

What we can have enabled by default is aforementioned
* Logging to stdout
* Prometheus endpoint
* profiling + tracing endpoint

And I see the appeal of easy integration into an existing Log Collection + Prometheus setup tuned for service discovery via Docker / k8s.

I suppose a good zrepl Docker container would
* auto-enable all these 'cloud-native' features (EXPOSE them on a port)
* and provide env vars to configure a *single* zrepl job (or maybe two just for local replication) following the idea that a single container exposes one service (provided by one job) or serves a single purpose (local replication)

Draft:

```
docker run --rm -it  \
    -e ZREPL_JOB=backup \
    -e ZREPL_FILESYSTEMS=<  \
    -e ZREPL_PUSH=tcp://10.2.5.10:3000 \
    -e ZREPL_SNAPSHOTTING="TODO_COMPACT_SYNTAX_FOR_SNAPSHOTTING" \
    -e ZREPL_PRUNING="TODO_COMPACT_SYNTAX_FOR_PRUNING" \
    zrepl 
```

I think we'd also benefit from just thinking about compact notation for things currently represented in YAML, e.g. `connect` specifications for each transport type, snapshotting + pruning syntax, etc.

enspritz:
Lots to consider.. Let us have me first derive a working Docker image.
`debian:stable-slim` produces an image of several hundred MiB, so I'm now looking at applying techniques to keep the distributable small.
Do you happen to know if zrepl will function correctly on Alpine?

problame:
I would like to reuse the existing `Makefile` to build the docker image:

- new Makefile target `docker-image`
- that target depends on `artifacts/zrepl-linux-amd64`
- The Dockerfile used to build the image simply copies artifacts/zrepl-linux-amd64 into a `scratch` container

enspritz:
Agreed. Amongst the patterns I've looked at here, this is approach should prove clean and integrative.
At the moment I'm playing with a multi-stage Dockerfile that expands on `build.Dockerfile`.

Please indicate a commit hash you prefer I start from; the obvious HEADs are reporting build errors for me.

enspritz:
Alrighty, I'm building off of master, bludgeoning errors by replacing offending functions with stubs for now. (which can have a real negative impact on testing the images.) And congratulations on releasing 0.3.0, it's better than ever before ^_^

Several questions, in the interest of reducing the build to its essential and necessary operations:

- golang build env requirements: `docs/installation/compile-from-source.rst` indicates Go 1.11 or newer, `build.Dockerfile` notes a new convenience introduced in Go 1.13 regarding modules, while `build.Dockerfile` uses 1.15. What is the actual state of affairs at this time?

- Does a running zrepl daemon need filesystem write access, other than to `/dev/zfs` and optionally user-configured logging? A read-only Docker image equates to an improvement in security. Even though an image runs `ro`, it can still require for example a tmpfs `/tmp`.

- How married are you to the current build process as documented in `docs/installation/compile-from-source.rst`. Lingering in my mind is the distinct possibility of authoring a Dockerfile build that produces the majority of your desired outcomes with a single issuance of `docker build`, kick-started by `make docker-image`.

- Relatedly, where & how are the various OS packages created, like `.deb`? 

- Is there a continuous integration system?

enspritz:
The change of substance to zrepl itself is that, in order to run in a `FROM scratch` environment, our lives are easier if `zrepl` is built statically. See the makefile for that change.

After building zrepl statically, we can now build and run the Docker image:
```bash
sd build --tag zrepl .
sd run --detach --name zrepl --read-only --device /dev/zfs --mount type=tmpfs,destination=/var/run/zrepl,tmpfs-size=1m zrepl
sd exec -it zrepl status    # Note: Must be interactive
sd logs zrepl               # See logs, use -f to view them live.
sd restart zrepl            # Restart the daemon
sd stop zrepl
```

With a nearly-blank configuration file as discussed above, zrepl will report in its logs:
```$bash
meine-maschine $ sudo docker logs zrepl
could not parse config: yaml: unmarshal errors:
  line 0: field jobs is required but not given
```

enspritz:
Thoughts at this point:

- Of all the YAML files spread throughout our fleet, zrepl is the only tool using the `.yml` file extension as opposed to `.yaml`. I felt a jarring sense of incongruence..

- The next step is to deploy the image in a safe capacity and observe it.

- Both `info` and `debug` levels of log output are satisfying so far.

- The pattern for running commands within the container, as written above, requires `zrepl` to be repeated. See if I can eliminate the need for that without losing anything of value..

problame:
> golang build env requirements: docs/installation/compile-from-source.rst indicates Go 1.11 or newer, build.Dockerfile notes a new convenience introduced in Go 1.13 regarding modules, while build.Dockerfile uses 1.15. What is the actual state of affairs at this time?

Go 1.11 is the minimum version required to build zrepl, the 1.13-thing in the build.Dockerfile is addressed by settings GO111MODULE=on in the makefile ( see https://dev.to/maelvls/why-is-go111module-everywhere-and-everything-about-go-modules-24k ).
The build.Dockerfile is what I use to produce the release binaries, and I want to build releases against the latest version of zrepl.

> 
Does a running zrepl daemon need filesystem write access, other than to /dev/zfs and optionally user-configured logging? A read-only Docker image equates to an improvement in security. Even though an image runs ro, it can still require for example a tmpfs /tmp.

AFAIK no, just the config, /dev/zfs, the `zfs` binary, the libraries that the `zfs` binary requires.
Thinking of it, the dependency on the `zfs` binary might be a showstopper for this effort, since the `zfs` binary must be the same or newer than the kmod version (there is some slack in the other way, but in general, they must stay in sync).

>  How married are you to the current build process as documented in docs/installation/compile-from-source.rst. Lingering in my mind is the distinct possibility of authoring a Dockerfile build that produces the majority of your desired outcomes with a single issuance of docker build, kick-started by make docker-image.
> 
>  Relatedly, where & how are the various OS packages created, like .deb?

It's what I use to build releases in a replicable fashion (I publish the Docker images I use to build the release binaries with every release).

Note that #369 will add packaging support for deb and rpm (`make rpm`, `make deb` respectively).
I believe that you should base your work on that, keep the CGO_ENABLED=0, and build the Docker image by copying-in the zrepl-linux-$arch binary into the scratch image, just like we do for rpm and deb in #369 

>  Is there a continuous integration system?

Yes, we use CircleCI. Look at the commits you pushed to this branch, they all have the little green checkmarks next to them.
#369 will integrate the rpm and deb creation into the CI.
We could do the same for the docker image eventually.

problame:
> Of all the YAML files spread throughout our fleet, zrepl is the only tool using the .yml file extension as opposed to .yaml. I continue to feel a jarring sense of incongruence..

Separate issue / direct PR please.

> The next step is to deploy the image in a safe capacity and observe its behavior.

Please play through the scenario of a version mismatch between kmod and zfs binary within the container.

enspritz:
> Yes, we use CircleCI.

I see it now. Thanks

> Thinking of it, the dependency on the zfs binary might be a showstopper for this effort, since the zfs binary must be the same or newer than the kmod version (there is some slack in the other way, but in general, they must stay in sync).

A capital point. The question is then: What compatibility guarantees does ZFS make here. And what is meant by "ZFS." Solaris/Illumos ZFS and/or OpenZFS? If zpools can be trashed due to version incompatibilities, what value remains in containerizing zrepl...

problame:
> If zpools can be trashed due to version incompatibilities, what value remains in containerizing zrepl...

The pools aren't likely to get trashed because the ioctls issued by the `zfs` binary inside the container will likely be rejected pretty soon by the zfs kmod's syscall handler.

I guess you'd need to find a way to proxy the `zfs` binary into the container, either through a stub shell script + ssh (bad performance guaranteed there) or by bind-mounting the `zfs` binary and its transitive closure of dependencies into the container.

The 'official' OpenZFS stance is that `libzfs_core` is de-facto stable, and that the CLI interface is stable.

InsanePrawn:
FWIW:
- Shipping/using the newest minor userspace utils for the right ZFS major version (0.8.4 utils for 0.8.x, 0.7.13 for 0.7.x) seems to be a thing and sounds sane enough to me: https://discuss.linuxcontainers.org/t/zfs-module-included-in-the-snap/4370/2
- Ansible, Symfony (and apparently Spring Boot) use `.yml` :) There's probably way more examples to be made. Also: https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml

enspritz:
> Shipping/using the newest minor userspace utils for the right ZFS major version

This might be sane. I'm trying it out. Resulting image size is 137MB.

> YAML

I'm not going to wade into this swamp!

problame:
@enspritz I just landed #369 into master. The `rpm` and `deb` targets in it re-use the artifacts produced by the `release` target.

With regards to the build system integration for this PR, I propose the following

* add a `make docker-image IMAGE=registry.example.com/zrepl/zrepl `
* that target should produce images with the following immutable tags that uniquely identifies a built image:
  `$(IMAGE):$(_ZREPL_VERSION)-$distro_codename-YYYYMMDD-HHMMSS`
* the CI must be adjusted to do
  * get docker-hub credentials (=> @zrepl-bot will do)
  * scheduled daily re-builds of the base images for the latest zrepl release (how do we teach CircleCI semantic versioning)
  * if the build is successful, move tags in the form `$(IMMAGE):$(_ZREPL_VERSION)-$distro_codename` forward to point to the new image

enspritz:
Any guidance on how you'd prefer zfs permissions to be arranged between host and container?

```
$ sudo docker exec -it zrepl sh
# zfs snapshot pool/...@zrepl_20200903_000000_000
cannot create snapshots : permission denied
```

problame:
Sry, I guess that's your job to figure out. Maybe the container needs to be privileged?

enspritz:
It seems that this Docker image approach to running zrepl is ready for more eyeballs.

Build the Docker image:
```bash
$ cd src/zrepl
$ sudo docker build --build-arg ZREPL_VERSION=0.3.0 --no-cache -t zrepl:0.3.0 .
```
Note that you can build the image with a clean checkout; no other build steps are necessary. zrepl is installed into the image using its official APT repo.

Allow zrepl running in the Docker image to perform necessary operations, for example:
```bash
$ sudo zfs allow root destroy,mount,snapshot tank
```
The `root` user specification is the user that the zrepl process runs as; confirm using `ps aux | grep zrepl` with a zrepl container running. Snapshots can be destroyed seemingly only with `mount` permissions. `allow`ed permissions are inherited by children; this scheme meshes well with zrepl's job `filesystems` spec.

Run the zrepl Docker image in a container:
```bash
sudo docker run --detach \
    --name zrepl \                     # A convenience for issuing commands using docker exec
    --read-only \
    --device /dev/zfs \                # zrepl requires access to this device file
    --mount type=tmpfs,destination=/run/zrepl,tmpfs-size=1m,tmpfs-mode=700 \
    -v /etc/hostid:/etc/hostid:ro \    # Convince zpool that the ZFS pools belong to it
    -v /etc/zrepl:/etc/zrepl:ro \      # Provide zrepl configuration file by mounting its dir as a volume in the container
    zrepl
```

enspritz:
> Maybe the container needs to be privileged?

TBH, I'm unaware of any methods other than `zfs allow`... Noted in the instructions above.

enspritz:
I have some results to share from running the Docker image in production:

- In the form of a Docker image, obtaining zrepl with `docker run` is easier that adding a package repo and installing from that. However, accommodating zrepl as a Docker image incurs some extra setup (as noted above). I feel this builds awareness in sysadmins regarding the bounds of zrepl's behavior.
- I didn't feel like I lost any capabilities with the Dockerized form; rather I felt liberated from systemd + journald & friends. If you are used of Docker.
- Our production environment includes a mix of nearly all x86-based types, including bare-metal Intel and AMD systems plus virtualised environments. All run zrepl. After changing from the zrepl daemon to the Docker image variant, we have unfortunately experienced several instances where a host would spontaneously freeze. No stack traces, no core dumps, no kernel logging, no apparent further common criteria between hosts. Although I would like to continue investing in attempting amelioration, the systems are too important to permit this phenomenon to occur again. If it helps, we are thinking towards ZFS and/or the Linux kernel as being responsible; zrepl is merely causing the problem to be expressed. But that could of course be wrong...

This last factor prevents me from being able to pursue Dockerized zrepl for the time being. I recommend closing this PR, either just dropping the issue for now, or incorporate generation of Docker images to the CI loop and push them to Docker hub or wherever with an alpha or beta tag.

problame:
> Our production environment includes a mix of nearly all x86-based types, including bare-metal Intel and AMD systems plus virtualised environments. All run zrepl. After changing from the zrepl daemon to the Docker image variant, we have unfortunately experienced several instances where a host would spontaneously freeze. No stack traces, no core dumps, no kernel logging, no apparent further common criteria between hosts. Although I would like to continue investing in attempting amelioration, the systems are too important to permit this phenomenon to occur again. If it helps, we are thinking towards ZFS and/or the Linux kernel as being responsible; zrepl is merely causing the problem to be expressed. But that could of course be wrong...

Yeah, zrepl only uses the ZFS CLI to interact with the kernel module, so if there is a deadlock, it's a bug in ZFS.

> This last factor prevents me from being able to pursue Dockerized zrepl for the time being. I recommend closing this PR, either just dropping the issue for now, or incorporate generation of Docker images to the CI loop and push them to Docker hub or wherever with an alpha or beta tag and let the community work out the problems over time.

I am still not really convinced that Dockerizing zrepl is a good idea due to its dependency on the `zfs` binary (and that binary's dependency on the loaded kernel module, which is outside of Docker's control).

My main take-away from this PR is: discovery of zrepl's Prometheus endpoint is simplified if it runs in Docker and there is already some auto-discovery mechanism for Prometheus endpoints exposed by docker containers.
I'm not that familiar with Prometheus service discovery - maybe zrepl can make things easier by integrating with some service discovery mechanism.
If you have ideas for that, please open an issue.

PR #359, opened 2020-08-22
Bump go build to v1.15

chenrui333:
kind of related to https://github.com/Homebrew/homebrew-core/pull/60011 (built via go1.15)

---

This PR depends on #360, #361


problame:
Thanks for this PR!

I have been contemplating about which Go versions be included in the CI and didn't come to a conclusion:

From my notes:

- Go 1.11 is the minimum Go version that zrepl needs (due to the dependency on problame/go-netssh).
- Go 1.13 added error wrapping. None of our dependencies nor zrepl itself use it ATM.
- Go 1.14 added preemption.
- Go 1.15 added the TLS SAN behavior.

I believe it makes sense to support roughly two years of Go such that
  - we avoid problems with downstream distro packagers (although one might argue that a distro stuck at a two year old version of Go will also be using a similarly old version of zrepl)
  - zrepl users on 'stable' distros with old Go versions in the distro-native package manager can compile zrepl from source (for debugging, etc). This is useful if
    - a user doesn't want to build in docker 
    - wants to get started with zrepl development and hasn't used Go before
    - is asked to apply a patch from a zrepl dev to debug an issue

That being said, I would have liked to start using the Go 1.13 error wrapping feature much sooner (see the `TODO go1.13`s in the zrepl code base).
And with the increasing pace of new Go features coming up, the 'roughly two years' might shift toward 'the second-to-latest Go release'.

With regard to pinning the Docker image `FROM`: Do you know a tool that will keep that `FROM` line up-to-date? I don't have experience with dependabot's Docker support, do you?

problame:
obsoleted by #369

chenrui333:
Thanks @problame!

PR #361, opened 2020-08-22
pIn dockerfile base image to v1.14.7

chenrui333:
relates to https://github.com/hadolint/hadolint/wiki/DL3007

> Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag.

### DIFF
--- a/build.Dockerfile
+++ b/build.Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:latest
+FROM golang:1.14.7
 
 RUN apt-get update && apt-get install -y \
     python3-pip \
@@ -27,4 +27,3 @@ ADD build/ /tmp/build
 RUN /tmp/lazy.sh godep
 
 RUN chmod -R 0777 /go
-


problame:
Go 1.15 has been released two weeks ago, so this PR is out-of-date and kind of serves an example for why we don't want to pin versions without periodic update checks.

@chenrui333 I just created https://github.com/zrepl/zrepl/issues/377 to comprehensively deal with Go version pinning.
I'd be interested in your opinion on my proposal there.

chenrui333:
> Go 1.15 has been released two weeks ago, so this PR is out-of-date and kind of serves an example for why we don't want to pin versions without periodic update checks.
> 
> @chenrui333 I just created #377 to comprehensively deal with Go version pinning.
> I'd be interested in your opinion on my proposal there.

@problame I actually think actively pinning the base image is trivial and serving the right context. Maybe it would be great idea to tune this into the changelog?

problame:
I'm not sure I understand what you mean by 'tune this into the changelog'.

The point I'm trying to make in #377 is that we shouldn't just lock the Go version to a specific version without a periodic check to see whether the most recent stable release of Go breaks something for us. We want to know about that early.

chenrui333:
yeah, dependabot is the way to go.

chenrui333:
what i was trying to say is outsourcing this maintenance overhead to the zrepl user (I believe community loves to help this out) and I do think pin down to a latest release is easy and trivial

problame:
> yeah, dependabot is the way to go.

Do you have experience with it's docker support? How does it determine if a newer docker image is available? Does it know about semver in image tags or does it just always use :latest?

Issue #367, opened 2020-08-27
zrepl v0.3.0 build fails on CentOS 7


poettler-ric:
Hello, I tried to update the packages for Fedora/CentOS. Everything except EPEL 7 builds fine. EPEL 7 (go version: 1.13.14) fails with the following message:
```
make ZREPL_VERSION=0.3.0 zrepl-bin
GO111MODULE=on go build -mod=readonly  -ldflags "-X github.com/zrepl/zrepl/version.zreplVersion=0.3.0" -o "artifacts/zrepl-linux-amd64"
go: gonum.org/v1/gonum@v0.7.0 requires
        gonum.org/v1/plot@v0.0.0-20190515093506-e2840ee46a6b requires
        github.com/fogleman/gg@v1.2.1-0.20190220221249-0403632d5b90: invalid pseudo-version: git fetch --unshallow -f origin in /builddir/go/pkg/mod/cache/vcs/68f39ed450731f4018ee308789f8e36cf8876a81d65be6a72eaa43bfa29c1a63: exit status 128:
        fatal: git fetch-pack: expected shallow list
```

If you let me know how you fixed that the last time (I am not the best go developer) I could send you a pull request the next time.

problame:
Did you have a look at #366 ? @awehrfritz put a new zrepl.spec together that seems to work. I wonder why though, since from what I can tell, their and you .spec look pretty much the same...

awehrfritz:
It seems the issue on RHEL 7 is an issue with an too old git version and Go 1.13 as indicated in these bug report: https://github.com/golang/go/issues/38373 and https://github.com/golang/go/issues/34092

The spec file referenced in https://github.com/zrepl/zrepl/issues/366 does not even consider RHEL 7 (i.e. it fails on a much earlier stage when configuring the build). I think we should first figure out a solution to this issue here, and can then consider adding the necessary pluming to the spec file in https://github.com/zrepl/zrepl/issues/366.

awehrfritz:
The issue here is indeed due to a too old git version in RHEL 7. Adding the [IUS community repositories](https://ius.io/) and requiring git >= 2 fixes the issue. I have implemented this in the spec file provided in PR #369.

poettler-ric:
So you can also add this repository during the build process on copr? Didn't know that.

problame:
RPM builds are addressed by #369 for zrepl 0.3.1 although.

@poettler-ric I won't issue 0.3.0 RPMs.
[...]

Issue #350, opened 2020-07-26
`make install` target


problame:
* should support PREFIX, see https://github.com/zrepl/zrepl/issues/325
   *use for correct systemd service definition

PR #369, opened 2020-08-30
build: rpm + deb targets, build-in-docker targets, CircleCI pipeline rewrite

### LINKED ISSUE #366: [Feature request] Provide packages/build instructions for the major Linux distributions
This is a feature request out of the frustration that has been piling up when trying to get `zrepl` installed on my Fedora Linux laptop as well as a Helios4 NAS running Debian Buster (with an Armbian ARMv7 kernel).

Little bit of background to elaborate on this request: The installation instruction for Fedora provided in the [docs](https://zrepl.github.io/installation/packages.html) refer to a Fedora [copr repo](https://copr.fedorainfracloud.org/coprs/poettlerric/zrepl/) which contains an outdated version. I then had a quick look at the spec file to update the package and realised that the spec file is more of a hack, required patches already for the outdated version, and would not work for the latest `zrepl` version. I then had an attempt to rewrite the spec from scratch following the latest Fedora [go packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Golang/), but I could not get it to work. Eventually, I looked at the [Makefile](https://github.com/zrepl/zrepl/blob/master/Makefile) (and [lazy.sh](https://github.com/zrepl/zrepl/blob/master/lazy.sh)) just to realise that this is all quite involved. I have no experience with Go, and thus cannot judge whether this is normal or not for Go, but I am working with with huge C++/Fortran projects (>1M lines of code) which are by far easier to build than this.

I understand that it is an additional effort to generate packages for downstream Linux distributions, but I believe it would make it at the same time much easier for the users to adopt the software. Also considering that a reasonably small Go-project like this really does not require much in terms of packaging, but apparently a lot of effort to get it compiled (simply judging by the convoluted Makefile). Furthermore, with services like Fedora's [copr](https://copr.fedorainfracloud.org/coprs/), openSUSE's [Open Build Service](https://build.opensuse.org/) and Ubuntu's [ppa](https://launchpad.net/ubuntu/+ppas) one can target already the vast majority of Linux users in an easy way (the .spec file for Fedora/RedHat and openSUSE/SUSE would probably even very similar).

I understand that `zrepl` itself is simply a binary and thus quite easy to distribute, but it comes in handy to have things like systemd service and bash-completion files installed in the appropriate places together with the binary itself, hence this feature request.

In the meantime, I resorted on my Fedora laptop to a simple shell script which pulls the binary and systemd-service file from GitHub and generates the bash-completion file on the fly. This is obviously not really how things should be done in the 21st-century, but it turns out that's the easiest way to get the latest `zrepl` version installed on Fedora.

awehrfritz:
This PR adds a spec file to generate rpm packages for Fedora, RedHat/CentOS >= 8 and openSUSE.

This has been discussed and partially resolves #366. The spec file has been used to generate rpm packages on Fedora's copr build service where packages for Fedora, RedHat/CentOS >= 8 and openSUSE are currently available in this project: [awehrfritz/zrepl](https://copr.fedorainfracloud.org/coprs/awehrfritz/zrepl/)

The modifications in the Makefile add build targets to generate srpm and rpm packages locally on a Fedora system, which will probably work on RedHat/CentOS >= 8 as well. This requires the respective rpm build tools to be available, which can be installed via:
``` sh
dnf install rpm-build rpmdevtools
```
Though it is probably a good idea to install the entire packaging toolchain as described in [quick-docs/creating-rpm-packages](https://docs.fedoraproject.org/en-US/quick-docs/creating-rpm-packages/) and [quick-docs/create-hello-world-rpm](https://docs.fedoraproject.org/en-US/quick-docs/create-hello-world-rpm/):
``` sh
dnf install fedora-packager
```

The rpm make targets will install the build dependencies and build the packages for the current host system. However, Fedora offers a simple way to build packages for other platforms in a chroot environment using `mock`, which avoid installing the build dependencies on the host system but requires the current user to be in the `mock` group.

Currently not covered in this PR are building rpm packages for the documentation or an integration into zrepl's CI. I suggest to address those in subsequent pull requests.

Also, once this is merged, I will change the Fedora copr repos to point to the spec file here.


awehrfritz:
I have pushed another commit to this branch which includes compatibility fixes for RHEL/CentOS 7. This thus resolves #367 as well.

problame:
Thanks for this contribution.
I just pushed my desired adjustments to the rpm build process to your branch.

  - use rpmbuild --build-in-tree to avoid creating the tarball
        - this also allows dirty builds, e.g. for development
        - as a consequence also avoid srpms
    
   - use a temporary _topdir within  the artifacts directory
        - that way the zrepl build's filesystem side-effects are limited to the checkout directory,
          which I, as a user, greatly appreciate
    
   - Drop the RPM changelog for now, I don't think I'd be maintaining it in future releases.

How do you feel about these changes? Are they compatible with copr?

I tested them both on my local Fedora 31 Workstation and with the following Dockerfile:

```
FROM fedora

RUN dnf install -y git make bash

RUN dnf install -y rpm-build rpmdevtools

RUN dnf install -y 'dnf-command(builddep)'

ADD packaging/rpm/zrepl.spec /tmp/zrepl.spec
RUN dnf builddep -y /tmp/zrepl.spec

WORKDIR /src
VOLUME /src
```

```
docker build -t zrepl_rpm_build  -f Dockerfile.rpmbuild . && docker run --rm -it -v `pwd`:/src zrepl_rpm_build bash
```

**My remaining critique** of this build process is that we re-build a zrepl binary within the process.
I'd much rather package the exact same binary that we ship for the release (fewer binaries out there => easier-to-reproduce user setups when they file issues).
What do you think about that?

problame:
@poettler-ric I'd be interested in your opinion as well!

awehrfritz:
> How do you feel about these changes? Are they compatible with copr?

I am fine with that, though this is drifting away from the original intention of this PR (i.e. adding a spec file for the downstream build systems). With these changes you cannot use the spec file on copr anymore. Also to note, my original spec file allows for dirty builds, simply do `make rpm` inside the repository. With these changes we don't need to worry anymore about compatibility with distribution specific things, we can simply aim at producing one rpm file for all distributions and versions. That's typically something distributions don't like, but clearly in favour of the upstream developers.


> **My remaining critique** of this build process is that we re-build a zrepl binary within the process.
> I'd much rather package the exact same binary that we ship for the release (fewer binaries out there => easier-to-reproduce user setups when they file issues).
> What do you think about that?

Actually, given the latest changes, I think it would make most sense to package the release binaries in the rpm-packages and simply host your own rpm repo. This would bypass the entire trouble of needing to maintain compatibility with the various distribution specific things, but would not really help in getting zrepl into the distributions directly (though that is probably difficult even with my original spec file due to the bundling). For most of it, hosting a own rpm repo requires simply a .repo file, something like https://copr.fedorainfracloud.org/coprs/awehrfritz/zrepl/repo/fedora-31/awehrfritz-zrepl-fedora-31.repo, or the ones Google adds for Chrome, or Microsoft for Skype, Teams and VScode.

Makefile
```diff
@@ -61,38 +61,33 @@ ifeq (SIGN, 1)
...
-else
-	$(eval _ZREPL_RPM_GIT_COMMIT := v$(_ZREPL_VERSION))
-	$(eval _ZREPL_RPM_VERSION := $(_ZREPL_VERSION))
-endif
-	rpmdev-setuptree
+	$(eval _ZREPL_RPM_VERSION := $(subst -,.,$(_ZREPL_VERSION)))
+	$(eval SRC_TARBALL_OUT := "$(HOME)/rpmbuild/SOURCES/zrepl-$(_ZREPL_RPM_VERSION).tar.gz")
+	$(eval _ZREPL_RPM_TOPDIR_ABS := $(PWD)/$(ARTIFACTDIR)/rpmbuild)
+	rm -rf "$(_ZREPL_RPM_TOPDIR_ABS)"
+	mkdir "$(_ZREPL_RPM_TOPDIR_ABS)"
```
awehrfritz, line 80: can be deleted

Makefile
```diff
@@ -61,38 +61,33 @@ ifeq (SIGN, 1)
...
-ifneq (,$(findstring -g,$(_ZREPL_VERSION)))
-	$(eval _ZREPL_RPM_GIT_COMMIT := HEAD)
-	$(eval _ZREPL_RPM_VERSION := g$(shell git rev-parse --short HEAD))
-else
-	$(eval _ZREPL_RPM_GIT_COMMIT := v$(_ZREPL_VERSION))
-	$(eval _ZREPL_RPM_VERSION := $(_ZREPL_VERSION))
-endif
-	rpmdev-setuptree
+	$(eval _ZREPL_RPM_VERSION := $(subst -,.,$(_ZREPL_VERSION)))
+	$(eval SRC_TARBALL_OUT := "$(HOME)/rpmbuild/SOURCES/zrepl-$(_ZREPL_RPM_VERSION).tar.gz")
```
awehrfritz, line 77: This directory is not guaranteed to be present as it is only created when running `rpmdev-setuptree`. Also, it voids your attempt to limit changes to the checkout directory. The original approach is altogether more elegant and easier to follow as everything related to a specific rpm build (be it for a release or development version) happens in a properly named sub-directory of ~/rpmbuild.

problame:
> Also to note, my original spec file allows for dirty builds, simply do make rpm inside the repository

That wasn't the case on my system: your code always did a git-archive call.

> With these changes you cannot use the spec file on copr anymore.

That's a bummer though, I don't want to maintain rpm repositories if it can be avoided.
(I'm also not fond of having to maintain the apt repos, but I do it because nobody else showed up).

> That's typically something distributions don't like, but clearly in favour of the upstream developers.

I don't think it makes sense to commit .spec or debian/rules intended for downstream distro packagers since they don't / shouldn't trust upstream devs with packaging anyways (afaik the package maintainer bastion is the whole point behind why people want a distro).

From my point of view, this PR was about integrating the creation of zrepl RPMs into the zrepl release process.

> Actually, given the latest changes, I think it would make most sense to package the release binaries in the rpm-packages and simply host your own rpm repo. This would bypass the entire trouble of needing to maintain compatibility with the various distribution specific things, but would not really help in getting zrepl into the distributions directly (though that is probably difficult even with my original spec file due to the bundling). For most of it, hosting a own rpm repo requires simply a .repo file, something like https://copr.fedorainfracloud.org/coprs/awehrfritz/zrepl/repo/fedora-31/awehrfritz-zrepl-fedora-31.repo, or the ones Google adds for Chrome, or Microsoft for Skype, Teams and VScode.

Huh, I should have read the entire comment first :D
How do you feel about packagecloud or similar services (please name them if you know any)?

---

**Side note**: I'm almost done with porting the .deb creation to the main zrepl repo. Would you mind me pushing them on top of this PR? I'd rename the PR to 'add .rpm and .deb packaging support to Makefile'

Makefile
problame: `SRC_TARBALL_OUT` was an oversight from an earlier experiment

problame: > The original approach is altogether more elegant and easier to follow as everything related to a specific rpm build (be it for a release or development version) happens in a properly named sub-directory of ~/rpmbuild.

Agreed, but at the expense of having to re-build zrepl (and the docs, which we haven't addressed yet, twice).
See https://github.com/zrepl/zrepl/pull/369#issuecomment-683444372

problame:
- [x] finished the transition of the .spec to just bundle up already-built artifacts.
- [x] added the equivalent deb packaging
- [x] added makefile target for deb
- [x] added makefile targets to build rpms and debs in Docker
- [ ] CircleCI integration
- [ ] automate DEB and RPM publishing to a repository (e.g. packagecloud.io)

I believe the only downside of this plan is that we couple a particular zrepl version/commit to a particular package release.
- pro: entire package building process in one repo, "easy" to understand and follow
- con: can't issue updated binary releases without bumping the zrepl version. When do we need this?
  - Go runtime security issue => Go project issues new update => zrepl needs to re-build with new Go release and issue an update
  - [ ] ideas how to address this?  how do other projects handle this?

@awehrfritz I apologize for basically taking over this PR, and maybe for misinterpreting your original intentions as well :/
Do you think RPM-ecosystem users would prefer a COPR repo over zrepl-provided RPM repos?

awehrfritz:
>> With these changes you cannot use the spec file on copr anymore.
>
>That's a bummer though, I don't want to maintain rpm repositories if it can be avoided.

Well, my original spec file was really targeted at Copr, it was neither fully compliant with the Fedora packaging guidelines, i.e. it would require justification of the bundling in order to get this into Fedora itself, nor was it really suited for building in-tree rpm packages. I think the way this evolved is better as this spec file is fully integrated into zrepl now. On a second thought, a spec file for Copr should probably be maintained outside the main zrepl repository due to the issues around incrementing the release counter when rebuilding.

>   * [ ]  automate DEB and RPM publishing to a repository (e.g. packagecloud.io)

Not having to host your own rpm repository is one of the perks of using Copr. But then again, setting up your own rpm repo similarly to how you handle the deb repository at the moment isn't too difficult, see e.g. the guide for CentOS: [HowTos/CreateLocalRepos](https://wiki.centos.org/HowTos/CreateLocalRepos). Packagecloud is an option indeed (though with a not insignificant price point), a free alternative is [jfrog.com/artifactory](https://jfrog.com/artifactory/) (assuming the monthly traffic is low enough).


> I believe the only downside of this plan is that we couple a particular zrepl version/commit to a particular package release.
>
> * pro: entire package building process in one repo, "easy" to understand and follow

I am all for that! :smile: 

> * con: can't issue updated binary releases without bumping the zrepl version. When do we need this?
>   
>   * Go runtime security issue => Go project issues new update => zrepl needs to re-build with new Go release and issue an update
>   * [ ]  ideas how to address this?  how do other projects handle this?

I don't see much of a problem in this, but I might be missing something. And given that the latest revision is pretty much independent of any downstream build system, there shouldn't be much of an issue with the rpm builds itself. It indeed evolved to become simply a packaging of the pre-built artifacts. Otherwise, I think there is much of a need for binary releases without having a zrepl version bump. Currently, with the standalone binaries there is no way to let a user know about an new build other than by a version bump which is communicated via GitHub or the like. For the rpm packages, the Release number became essentially useless, but then again, it remains consistent with the standalone binary release.


> @awehrfritz I apologize for basically taking over this PR, and maybe for misinterpreting your original intentions as well :/

That's alright. After all, I mostly care about having and consistent workflow for package building, and it is even better if that is integrated in the release process.


> Do you think RPM-ecosystem users would prefer a COPR repo over zrepl-provided RPM repos?

From a user perspective, I think Copr repos provided by some third party (like myself) are less favourable than a rpm repo provided by the upstream project. (There really aren't any checks on what is built/packaged in a Copr repo.) I generally prefer distribution packages (if they are reasonably new) as they are indeed vetted and tested by typically trustworthy people, or otherwise packages maintained by the upstream developers (my prime example for this would be the [Intel OpenCL project on Copr](https://copr.fedorainfracloud.org/coprs/jdanecki/intel-opencl/)).

packaging/rpm/zrepl.spec
```diff
@@ -0,0 +1,100 @@
...
+
+# Add workaround for "No build ID note found" error during %%install
+%undefine _missing_build_ids_terminate_build
+
+
+# Disable debuginfo for RedHat/CentOS builds to avoid errors due to empty
+# debuginfo files, see also: https://fedoraproject.org/wiki/Packaging:Debuginfo
+%if 0%{?rhel}
+%global debug_package %{nil}
+%endif
```
awehrfritz, line 14: Disable the debug packages everywhere, which further allows to remove the workaround for the build ID issue:
```suggestion
# Disable debug packages as the binaries are not built by this spec file, see also:
# https://fedoraproject.org/wiki/Packaging:Debuginfo
￼%global debug_package %{nil}
```

packaging/rpm/zrepl.spec
```diff
@@ -0,0 +1,100 @@
...
+zrepl is a one-stop, integrated solution for ZFS replication.}
+
+
+Name:             zrepl
+Release:          1%{?dist}
+Summary:          One-stop, integrated solution for ZFS replication
+License:          MIT
+URL:              https://zrepl.github.io/
+# Source: we use rpmbuild --build-in-tree => no source
+BuildRequires:    git >= 2
```
awehrfritz, line 29:
```suggestion
```
Not needed when packaging binaries.

docs/installation/packages.rst
```diff
@@ -30,12 +30,15 @@ The following list may be incomplete, feel free to submit a PR with an update:
...
-      - Available on `COPR <https://copr.fedorainfracloud.org/coprs/poettlerric/zrepl/>`_
-    * - CentOS/RHEL
+      - Available on `Fedora Copr <https://copr.fedorainfracloud.org/coprs/awehrfritz/zrepl/>`_
+    * - CentOS 7
       - ``yum install zrepl``
-      - Available on `COPR <https://copr.fedorainfracloud.org/coprs/poettlerric/zrepl/>`_
+      - Available on `Fedora Copr <https://copr.fedorainfracloud.org/coprs/awehrfritz/zrepl/>`_
+    * - openSUSE
+      - ``zypper install zrepl``
+      - Available on `Fedora Copr <https://copr.fedorainfracloud.org/coprs/awehrfritz/zrepl/>`_
```
awehrfritz, line 41: This is good for now (i.e. version 0.3.0) as my Copr project is still live and in fact uses the original spec file from a separate git repo, but eventually this should be updated to point to the new rpm repo (with instructions on how to add that new repo).

packaging/rpm/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM fedora:latest
+
+RUN dnf install -y git make bash rpm-build rpmdevtools 'dnf-command(builddep)'
```
awehrfritz, line 3:
```suggestion
RUN dnf install -y git make bash rpm-build 'dnf-command(builddep)'
```
I haven't tested it, but I think you can get rid of `rpmdevtools` since you removed `rpmdev-setuptree` from the Makefile. If needed it will be pulled in later by `dnf builddep`.

packaging/rpm/zrepl.spec
```diff
@@ -0,0 +1,100 @@
...
+
+%install
+install -Dm 0755 artifacts/%{_zrepl_binary_filename}  %{buildroot}%{_bindir}/zrepl
+install -Dm 0644 artifacts/rpmbuild/zrepl.service  %{buildroot}%{_unitdir}/zrepl.service
+install -Dm 0644 artifacts/_zrepl.zsh_completion      %{buildroot}%{_datadir}/zsh/site-functions/_zrepl
+install -Dm 0644 artifacts/bash_completion     %{buildroot}%{_datadir}/bash-completion/completions/zrepl
+install -Dm 0644 artifacts/rpmbuild/zrepl.yml      %{buildroot}%{_sysconfdir}/zrepl/zrepl.yml
+install -d %{buildroot}%{_datadir}/doc/zrepl
+cp -a 	artifacts/docs/html		%{buildroot}%{_datadir}/doc/zrepl/html
+cp -a 	config/samples       %{buildroot}%{_datadir}/doc/zrepl/examples
```
awehrfritz, line 73: The space between source and destination in the install commands is all messed up now. Either make it one space only or align all destinations for enhanced readability.

Makefile
```diff
@@ -57,6 +61,77 @@ ifeq (SIGN, 1)
...
+	$(eval _ZREPL_RPMBUILD_TARGET := $(GOARCH))
+endif
+	rpmbuild \
+		--build-in-place \
+	  --define "_sourcedir $(CURDIR)" \
+	  --define "_topdir $(_ZREPL_RPM_TOPDIR_ABS)" \
+	  --define "_zrepl_binary_filename zrepl-$(ZREPL_TARGET_TUPLE)" \
+	  --target $(_ZREPL_RPMBUILD_TARGET) \
+	  -bb "$(_ZREPL_RPM_TOPDIR_ABS)"/SPECS/zrepl.spec
+	cp "$(_ZREPL_RPM_TOPDIR_ABS)"/RPMS/*/zrepl-$(_ZREPL_RPM_VERSION)*.rpm $(ARTIFACTDIR)/
```
awehrfritz, line 101:
```suggestion
	cp "$(_ZREPL_RPM_TOPDIR_ABS)"/RPMS/$(_ZREPL_RPMBUILD_TARGET)/zrepl-$(_ZREPL_RPM_VERSION)*.rpm $(ARTIFACTDIR)/
```

Makefile
```diff
@@ -57,6 +61,77 @@ ifeq (SIGN, 1)
...
+else
+	$(eval _ZREPL_RPMBUILD_TARGET := $(GOARCH))
+endif
+	rpmbuild \
+		--build-in-place \
+	  --define "_sourcedir $(CURDIR)" \
+	  --define "_topdir $(_ZREPL_RPM_TOPDIR_ABS)" \
+	  --define "_zrepl_binary_filename zrepl-$(ZREPL_TARGET_TUPLE)" \
+	  --target $(_ZREPL_RPMBUILD_TARGET) \
+	  -bb "$(_ZREPL_RPM_TOPDIR_ABS)"/SPECS/zrepl.spec
```
awehrfritz, line 100: make the indentation consistent

Makefile
```diff
@@ -57,6 +61,77 @@ ifeq (SIGN, 1)
 endif
 	@echo "ZREPL RELEASE ARTIFACTS AVAILABLE IN artifacts/release"
 
+rpm: $(ARTIFACTDIR) # artifacts/_zrepl.zsh_completion artifacts/bash_completion docs zrepl-bin
```
awehrfritz, line 64: Not quite sure how this is working as a `make rpm` in a clean repository should fail due to missing the zrepl binary. Or am I missing something here?

packaging/rpm/zrepl.spec
```diff
@@ -0,0 +1,100 @@
...
+	      format: human
+	      level: warn
+
+	jobs:
+	#  - name: foo
+	#    type: bar
+
+	# see %{_datadir}/doc/zrepl/examples
+	# or https://zrepl.github.io/configuration/overview.html
+EOF
```
awehrfritz, line 62: Given that the spec file is not standalone anymore anyway, the default config file should be added to the git repository and copied (with replacing the path to the examples using sed) analog to the systemd file. This would ensure that the same config file is distributed in all packages (rpm, deb, ...).

packaging/rpm/zrepl.spec
```diff
@@ -0,0 +1,100 @@
...
+BuildRequires:    git >= 2
+BuildRequires:    systemd
+Requires(post):   systemd
+Requires(preun):  systemd
+Requires(postun): systemd
+
+%description
+%{common_description}
+
+$prep
```
awehrfritz, line 38:
```suggestion
%prep
```

poettler-ric:
> @poettler-ric I'd be interested in your opinion as well!

What I know for a fact is that CentOS 7 always caused headaches when it came to packaging (explains the hack). Leaving out CentOS 7 is not an option due to it's wide application.
If you can ship around those issues by using an additional repository with the trade off of not using copr, I would do it. The cost for doing this is finding ways to build all needed packages (seems to be docker, for what I've seen so far (never used it)) and finding a place to host the repository. I also don't have a clue on how to automatically sign a package.

Do I think continuously building the packages to check everything is still in line will lead to better quality? Yes.

problame:
> Not having to host your own rpm repository is one of the perks of using Copr. But then again, setting up your own rpm repo similarly to how you handle the deb repository at the moment isn't too difficult, see e.g. the guide for CentOS: HowTos/CreateLocalRepos. Packagecloud is an option indeed (though with a not insignificant price point), a free alternative is jfrog.com/artifactory (assuming the monthly traffic is low enough).

I took a look at `createrepo` + `rpmsign` and have decided that I'll run the rpm repos myself as well.
AFAICT it's possible to just dump all rpms for multiple zrepl versions and multiple archs into the same repository, and dnf will just do the right thing. Is that correct?

> 
I don't see much of a problem in this, but I might be missing something. And given that the latest revision is pretty much independent of any downstream build system, there shouldn't be much of an issue with the rpm builds itself. It indeed evolved to become simply a packaging of the pre-built artifacts. Otherwise, I think there is much of a need for binary releases without having a zrepl version bump. Currently, with the standalone binaries there is no way to let a user know about an new build other than by a version bump which is communicated via GitHub or the like. For the rpm packages, the Release number became essentially useless, but then again, it remains consistent with the standalone binary release.

Right, it's the same problem for raw binary releases, rpms and debs alike. I'm willing to defer addressing a procedure for that for a later time. I created #374 for that.

docs/installation/packages.rst
problame: My plan is to offer the RPM repo as of 0.3.1 which I plan to release in a few weeks at most.

Makefile
problame: Why are you proposing this? The fedora 32 container separates creates one subdirectory per architecture.

awehrfritz: Because we know the value of this for sure now and adding it explicitly will help to prevent unforeseen issues due to a unnecessary wildcard - simply a safe-by-design choice.

awehrfritz:
> I took a look at `createrepo` + `rpmsign` and have decided that I'll run the rpm repos myself as well.
> AFAICT it's possible to just dump all rpms for multiple zrepl versions and multiple archs into the same repository, and dnf will just do the right thing. Is that correct?

I believe this should be sufficient, but I have only setup local repositories (i.e. on the local system) myself so far and never a proper publicly available one.

Makefile
problame: Nope, the general problem with the Makefile is that most of the targets (including zrepl-bin, for example) are phony (i.e. execute their commands every time). Very much so in this section of the makefile that packages up output from other make targets. Look at the `release` target for example, it's basically a shell script. I think of `rpm` and `deb` in a similar way.

This situation is not great but I can't find a maintainable solution to this ATM.

problame:
@awehrfritz @poettler-ric I have set up an RPM repository with a development build of zrepl in it.

Could you please test the repo using the following docs, ideally on both x86_64 and another architecture?
Also, please try older rpm versions, e.g. CentOS 7 / OpenSUSE $OLD_VERSION.
I already switched the zrepl.spec to gzip compression in order to avoid incompatibilties due to zstd usage on fc32

```
.. _installation-rpm-repos:

RPM repositories
~~~~~~~~~~~~~~~~

We provide a single RPM repository for all RPM-based Linux distros.
The zrepl binary in the repo is the same as the one published to GitHub.
Since Go binaries are statically linked, the RPM should work about everywhere.

The fingerprint of the signing key is ``F6F6 E8EA 6F2F 1462 2878 B5DE 50E3 4417 826E 2CE6``.
It is available at `<https://zrepl.cschwarz.com/rpm/rpm-key.asc>`_ .
Please open an issue on GitHub if you encounter any issues with the repository.

Copy-paste the following snippet into your shell to set up the zrepl repository.
Then ``dnf install zrepl`` and make sure to confirm that the signing key matches the one shown above.

.. literalinclude:: ../../packaging/rpm/zrepl.repo
::

    cat > /etc/yum.repos.d/zrepl.repo <<EOF
    [zrepl]
    name = zrepl
    baseurl = https://zrepl.cschwarz.com/rpm/repo
    gpgkey = https://zrepl.cschwarz.com/rpm/rpm-key.asc
    EOF

.. NOTE::

   Until zrepl reaches 1.0, the repository will be updated to the latest zrepl release immediately.
   This includes breaking changes between zrepl versions.
   If that bothers you, use the `dnf versionlock plugin <https://dnf-plugins-core.readthedocs.io/en/latest/versionlock.html>`_ to pin the version of zrepl on your system.
```

problame:
- [x] repository works on AWS CentOS 7 x86_64
- [x] repository works on Fedora 31 x86_64
- [x]  repository works on Amazon Linux on Arm 64bit (aarch64)
- [x] interop test between aws centos 7 x86_64 and arm 64bit

There are systemd warnings/errors on old CentOS versions because we use newer systemd features.
Systemd starts the service anyway and simply ignores the newer config fields.
That's good enough for this PR to be merged.
#350 (`make install` target should probably address systemd service definition generation as well)

problame:
Merged!

@awehrfritz thanks again for your initiative and my apologies again for taking over this PR.
Starting with the 0.3.1 release (I'm planning to release it in a few weeks at most), official zrepl RPM repositories are going to be available.

awehrfritz:
Great!

I'm happy with how this whole thing turned out! Well documented and relatively easy to follow package-build process for probably all rpm and deb based distributions - pretty much what I was hoping for!

This should make it much easier for many new users to get started, and the maintenance burden isn't too big either.

Thanks @problame for all your hard work with zrepl and for provide such a neat and useful utility!

PR #372, opened 2020-08-31
lazy.sh: various fixes

InsanePrawn:
- Emit warning and return 1 when no cmd is specified
- Explicitly warn when gopath is not in PATH
- get docdep requirements from lazy.sh parent dir, not $GOPATH/src/zrepl/zrepl
- Fix `release` cmd
- replace `! -z` with `-n`

### DIFF
--- a/lazy.sh
+++ b/lazy.sh
@@ -20,11 +20,14 @@ if ! type go >/dev/null; then
 fi
 
 if [ -z "$GOPATH" ]; then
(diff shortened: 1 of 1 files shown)


lazy.sh
```diff
@@ -63,9 +67,15 @@ release() {
     make release
 }
 
+    # shellcheck disable=SC2198
+if [ -z "$@" ]; then
```
problame, line 71: indentation

problame (approved):

PR #391, opened 2020-10-07
Fixing link to template.sh sample hook file

helamonster:
The link to the sample hook file currently goes to sampleshooks/template.sh instead of samples/hooks/template.sh


problame:
Thanks!

Issue #302, opened 2020-04-07
go race detector detects data races within in rpc stack with parallel replication enabled


problame:
This one seems somewhat legit:
```
WARNING: DATA RACE
Write at 0x00c00015fe50 by goroutine 26:
  sync/atomic.CompareAndSwapInt32()
      /home/cs/go1.13.0/src/runtime/race_amd64.s:293 +0xb
  github.com/zrepl/zrepl/rpc/dataconn/timeoutconn.(*Conn).DisableTimeouts()
      /home/cs/zrepl/zrepl/rpc/dataconn/timeoutconn/timeoutconn.go:69 +0x51
  github.com/zrepl/zrepl/rpc/dataconn/frameconn.(*Conn).Shutdown()
      /home/cs/zrepl/zrepl/rpc/dataconn/frameconn/frameconn.go:329 +0x15a
  github.com/zrepl/zrepl/rpc/dataconn/heartbeatconn.(*Conn).Shutdown()
      /home/cs/zrepl/zrepl/rpc/dataconn/heartbeatconn/heartbeatconn.go:72 +0x11c
  github.com/zrepl/zrepl/rpc/dataconn/stream.(*Conn).Close()
      /home/cs/zrepl/zrepl/rpc/dataconn/stream/stream_conn.go:298 +0x91
  github.com/zrepl/zrepl/rpc/dataconn.(*Client).putWire()
      /home/cs/zrepl/zrepl/rpc/dataconn/dataconn_client.go:107 +0x3c
  github.com/zrepl/zrepl/rpc/dataconn.(*Client).ReqRecv()
      /home/cs/zrepl/zrepl/rpc/dataconn/dataconn_client.go:204 +0x98e
  github.com/zrepl/zrepl/rpc.(*Client).Receive()
      /home/cs/zrepl/zrepl/rpc/rpc_client.go:100 +0x8d
  github.com/zrepl/zrepl/replication/logic.(*Step).doReplication()
      /home/cs/zrepl/zrepl/replication/logic/replication_logic.go:636 +0x563
  github.com/zrepl/zrepl/replication/logic.(*Step).Step()
      /home/cs/zrepl/zrepl/replication/logic/replication_logic.go:186 +0x4c
  github.com/zrepl/zrepl/replication/driver.(*fs).do.func5()
      /home/cs/zrepl/zrepl/replication/driver/replication_driver.go:565 +0x1b9
  github.com/zrepl/zrepl/util/chainlock.(*L).DropWhile()
[...]

PR #413, opened 2021-01-23
rpc: fix read mutex unlock in dataconn stream

mafredri:
I took a closer look at the goroutine traces in #411 to narrow down the scope and found one suspicious mutex that was not being unlocked in the event of an error (unclean read).

Currently putting this PR through the test to see if it fixes the issues I've been having. May potentially fix the freebsd side as well?

### DIFF
--- a/rpc/dataconn/stream/stream_conn.go
+++ b/rpc/dataconn/stream/stream_conn.go
@@ -144,6 +144,7 @@ func (c *Conn) ReadStream(frameType uint32, closeConnOnClose bool) (_ *StreamRea
 
 	c.readMtx.Lock()
 	if !c.readClean {
+		c.readMtx.Unlock()
 		return nil, errWriteStreamToErrorUnknownState
 	}
 


rpc/dataconn/stream/stream_conn.go
```diff
@@ -143,13 +143,13 @@ func (c *Conn) ReadStream(frameType uint32, closeConnOnClose bool) (_ *StreamRea
 	}
 
 	c.readMtx.Lock()
+	defer c.readMtx.Unlock()
```
problame, line 146: This unlocks the mutex too early in the case where `c.readClean == true`: the idea is that the goroutine spawned below copies the stream into the io.Pipe and the returned StreamReader reads from it. The Conn is not usable until StreamReader is drained.

So I think we just need to leave the `defer unlock` in the goroutine as it was but add a non-defer unlock in the erorr case where c.readClean == false.

problame:
Thanks for looking into this! CircleCI is apparently having a maintenance window today, that's why you don't get CI results. But there aren't any tests for this code path anyways.

problame:
While you are at it also have a look at https://github.com/zrepl/zrepl/issues/302

rpc/dataconn/stream/stream_conn.go
mafredri: Ah, I glanced too quickly and missed the purpose of the defer placement. That was actually my first solution but ended moving the defer, I’ll update the PR 👍

problame:

PR #414, opened 2021-01-24
daemon: guard trace id generation with mutex

mafredri:
This is probably low-hanging fruit but was caught by the race detector and was an easy fix.

Let me know if the go lint fix of `Id` -> `ID` is unwanted and I'll revert.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"os"
 	"os/signal"
 	"strings"
@@ -37,6 +38,12 @@ func Run(ctx context.Context, conf *config.Config) error {
 		<-sigChan
 		cancel()
 	}()
+
+	// The math/rand package is used presently for generating trace IDs, we
+	// seed it with the current time and pid so that the IDs are mostly
+	// unique.
+	rand.Seed(time.Now().UnixNano())
+	rand.Seed(int64(os.Getpid()))
 
 	outlets, err := logging.OutletsFromConfig(*conf.Global.Logging)
 	if err != nil {
--- a/daemon/logging/trace/trace_genID.go
+++ b/daemon/logging/trace/trace_genID.go
@@ -3,19 +3,10 @@ package trace
 import (
 	"encoding/base64"
 	"math/rand"
-	"os"
 	"strings"
-	"time"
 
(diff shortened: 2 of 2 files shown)


mafredri:
I just noticed the trace posted here https://github.com/zrepl/zrepl/issues/411#issuecomment-765586935; and there's a chance this PR could fix it.

problame:
> The default Source is safe for concurrent use by multiple goroutines, but Sources created by NewSource are not. 

https://golang.org/pkg/math/rand/

We are not using the default source, but `var genIdPRNG = rand.New(rand.NewSource(1))` instead. I can't remember whether this was intentional or an oversight. But couldn't we just use the default source and let package `math/rand` about concurrency?

mafredri:
> We are not using the default source, but `var genIdPRNG = rand.New(rand.NewSource(1))` instead. I can't remember whether this was intentional or an oversight. But couldn't we just use the default source and let package `math/rand` about concurrency?

I don't see why not. We'd want to think about how `math/rand` is used throughout the application though and how/if to seed it. I usually like to seed it (where seeding is required) in `func main()` to avoid conflicts between packages.

The only other (non test) use seems to be `zrepl/rpc/dataconn/heartbeatconn/integration_test_variablereceiverate/heartbeatconn_integration_variablereceiverate.go:113` though, so I don't think we'll run into any issues. But seeding based on `time.Now()` is probably good to ensure we receive unique IDs for the traces.

problame:
> The only other (non test) use seems to be zrepl/rpc/dataconn/heartbeatconn/integration_test_variablereceiverate/heartbeatconn_integration_variablereceiverate.go:113 though, so I don't think we'll run into any issues. But seeding based on time.Now() is probably good to ensure we receive unique IDs for the traces. (Edit: I just realized the above is also a test package, heh.)

That's an integration test (which we shamefully don't even build nor run in CI (created https://github.com/zrepl/zrepl/issues/417for this ) so it's its own `package main` anyways. Seeding from `main()` seems reasonable.

Would you do the change and force-push over this branch?

mafredri:
Sure, I should be back at my computer in about an hour, can do the change then 👍

mafredri:
I had a look at `func main()` and since it was only one row, I opted to do `rand.Seed` in `daemon.Run`. Also, since the scope changed, I undid the Id -> ID go lint change.

problame:
Thanks!

problame:

PR #415, opened 2021-01-24
rpc: fix data race in timeoutconn

mafredri:
I took a look at #302 and while I can't explain the the full story behind that data race, I did find a potential source for the issue.

Thought process:

- `timeoutconn` handles state, yet calls to Read/Write make a copy of that state (non-pointer receiver) so any outbound calls will not have the state updated
- Even without the copy issue, the renew methods can in edge cases set a new deadline _after_ DisableTimeouts have been called, consider the following racy behavior:
    1. `renewReadDeadline` is called, checks `renewDeadlinesDisabled` (not disabled)
    2. `DisableTimeouts` is called, sets `renewDeadlinesDisabled`
    3. `DisableTimeouts` invokes `c.SetDeadline`
    4. `renewReadDeadline` invokes `c.SetReadDeadline`

To fix the above, `Conn` was made into a pointer and `atomic` replaced with a `mutex`.

### DIFF
--- a/rpc/dataconn/timeoutconn/timeoutconn.go
+++ b/rpc/dataconn/timeoutconn/timeoutconn.go
@@ -68,7 +68,8 @@ func Wrap(conn Wire, idleTimeout time.Duration) *Conn {
 }
 
 // DisableTimeouts disables the idle timeout behavior provided by this package.
-// Existing deadlines are cleared iff the call is the first call to this method.
+// Existing deadlines are cleared iff the call is the first call to this method
+// or if the previous call produced an error.
 func (c *Conn) DisableTimeouts() error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
--- a/rpc/dataconn/timeoutconn/timeoutconn.go
+++ b/rpc/dataconn/timeoutconn/timeoutconn.go
@@ -72,10 +72,14 @@ func Wrap(conn Wire, idleTimeout time.Duration) *Conn {
 func (c *Conn) DisableTimeouts() error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	if !c.renewDeadlinesDisabled {
-		c.renewDeadlinesDisabled = true
-		return c.SetDeadline(time.Time{})
+	if c.renewDeadlinesDisabled {
+		return nil
 	}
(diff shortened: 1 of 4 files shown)


mafredri:
Hmm, is there any way to view the circleci logs _without_ giving them access to my public _and_ private repositories? (I don't have an account there and can't seem to view the log without one.)

mafredri:
I've been running this with zrepl built with `-race` and `ZREPL_REPLICATION_EXPERIMENTAL_REPLICATION_CONCURRENCY=4` set. Before I was able to reproduce the data race from #302 immediately, but not after running this PR for 2-3h. Please note this has been running with my previous 2 PRs merged in for a while and afterwards with #416 as well.

problame:
> timeoutconn handles state, yet calls to Read/Write make a copy of that state (non-pointer receiver) so any outbound calls will not have the state updated

good catch!

>     Even without the copy issue, the renew methods can in edge cases set a new deadline after DisableTimeouts have been called, consider the following racy behavior:
        renewReadDeadline is called, checks renewDeadlinesDisabled (not disabled)
        DisableTimeouts is called, sets renewDeadlinesDisabled
        DisableTimeouts invokes c.SetDeadline
        renewReadDeadline invokes c.SetReadDeadline


Also a very good catch :)

rpc/dataconn/timeoutconn/timeoutconn.go
```diff
@@ -73,9 +73,9 @@ func (c *Conn) DisableTimeouts() error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	if !c.renewDeadlinesDisabled {
+		c.renewDeadlinesDisabled = true
```
problame, line 76: Hm, shouldn't we do this only if the c.SetDeadline call below doesn't return an error?

mafredri: I agree, that makes the most sense 👍🏻

problame:
I'll squash all the commits into one and extend the commit message with contents from this PR, ok?

mafredri:
> I'll squash all the commits into one and extend the commit message with contents from this PR, ok?

Please do, that's my preferred way as well. :)

problame:

PR #420, opened 2021-02-13
Systemd: Add zrepl-job@.{timer,service} units and zrepl_timers.target

InsanePrawn:
Allows a user to trigger a job through systemd by starting zrepl-job@`$JOBNAME|systemd-escape`.service,
e.g. by simply enabling zrepl-job@`$JOBNAME|systemd-escape`.timer and setting the desired trigger condition,
e.g. via systemctl edit zrepl-job@`$JOBNAME|systemd-escape`.timer.

## TODO

- [ ] Documentation
- [ ] Finalize unit names (zrepl_timers.target? zrepl-timers.target?)
- [ ] Wait for 0.4 and new signalling code?
- [ ] Decide on what type of subshell-ish syntax to use for the pseudocode in the commit message 🙃

### DIFF
--- a/dist/systemd/zrepl-job.service
+++ b/dist/systemd/zrepl-job.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=zrepl job %i wakeup
+After=zrepl.service
+
+# Don't do this to yourself
+# unless you want job units to keep starting zrepl while you're trying to keep the daemon stopped for a moment...
+# Requires=zrepl.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/zrepl signal wakeup %i
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


problame:
Why is the timer target necessary?

InsanePrawn:
> Why is the timer target necessary?

An excellent question to which I don't have any answer. If someone else has a grounded opinion on whether it should stay or be replaced (with what? timers.target?), I'd be glad :)

problame:
I guess use cases [like this one](https://zrepl.github.io/quickstart/backup_to_external_disk.html) require cron or a timer.

But IMO it doesn't make sense to ship a oneshot service + corresponding timer for that.
It adds more confusion than is worth it.
We should rather work to eliminate the need for running `zrepl signal wakeup` from cron.
(I know, we have various issues for this, first and foremost https://github.com/zrepl/zrepl/pull/462#issuecomment-821842693 . Hoping to get back to that soon.)

PR #436, opened 2021-03-04
[WIP] Add "wait" command for blocking until a job is finished

### LINKED ISSUE #427: Allow waiting for a woken-up job to finish
I'm using the [Local Snapshots + Offline Backup to an External Disk example](https://zrepl.github.io/quickstart/backup_to_external_disk.html) to backup a local pool to an external (i.e. **detachable**) disk.

I've created a script to decrypt the LUKS container on it, import the pool, wake up the push job ... well here my setup gets ugly: I've found no way to wait for the job I've just woken up to finish so I can export the pool and close the LUKS container cleanly.

Similar to [others]( https://github.com/zrepl/zrepl/issues/245#issuecomment-556016281) I also do brittle polling of `zrepl status --raw | jq 'magic'` in a loop until everything is done or something fails.

My feature suggestion:
Add a `-w` or `--wait` flag to `zrepl signal wakeup` that blocks until the job is done or something failed (exiting with non-zero code preferably :wink:). Alternatively there could be a `zrepl wait ...` subcommand.
FWIW: this would mimic behavior also present in OpenZFS. They also added `-w` flags to several long-running tasks as well as a [`zpool wait` command](https://openzfs.github.io/openzfs-docs/man/8/zpool-wait.8.html) .

riyad:
This is a rough first implementation for  #427.

Design decisions:
- calling the subcommand "wait"
- allow waiting only for one job

Non-descisions (i.e. there was no thought behind it):
- code structure and variable names mimic client/status
- sleep time between checks (currently 500msec)

Caveats:
- I'm writing Go for the first time
- I have relatively little experience with zrepl
- I only have a simple setup
- no (integration) tests

Todos:
- cover more corner cases ...
  -  .. through feedback 
  -  ... with tests
- better naming and code structure (any suggestions guidelines?)
- should there be meaningful error codes?
- should there be status/debug output?

Fixes  #427

### DIFF
--- a/client/wait.go
+++ b/client/wait.go
@@ -1,0 +1,276 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/gdamore/tcell/termbox"
+	"github.com/pkg/errors"
+
+	"github.com/zrepl/zrepl/cli"
+	"github.com/zrepl/zrepl/config"
+	"github.com/zrepl/zrepl/daemon"
+	"github.com/zrepl/zrepl/daemon/job"
+	"github.com/zrepl/zrepl/daemon/pruner"
+	"github.com/zrepl/zrepl/daemon/snapper"
+	"github.com/zrepl/zrepl/replication/report"
+)
+
+type wui struct {
+	//lock      sync.Mutex //For report and error
+	report map[string]*job.Status
+	cancel bool
+
+	jobName string
+}
+
+var WaitCmd = &cli.Subcommand{
+	Use:   "wait JOB",
+	Short: "block and wait until the specified job is done or has failed",
+	Run: func(ctx context.Context, subcommand *cli.Subcommand, args []string) error {
+		return runWaitCmd(subcommand.Config(), args)
+	},
+}
+
+// adapted from status.go
+func runWaitCmd(config *config.Config, args []string) error {
+	if len(args) != 1 {
+		return errors.Errorf("Expected 1 argument: JOB")
+	}
+
+	// TODO: poll status until done or error
+	httpc, err := controlHttpClient(config.Global.Control.SockPath)
+	if err != nil {
+		return err
+	}
+
+	t := wui{
+		report:  nil,
+		cancel:  false,
+		jobName: args[0],
+	}
+
+	// wait for cancellation by user
+	err = termbox.Init()
+	if err != nil {
+		return err
+	}
+	defer termbox.Close()
+	go func() {
+	loop:
+		for {
+			switch ev := termbox.PollEvent(); ev.Type {
+			case termbox.EventKey:
+				switch ev.Key {
+				case termbox.KeyCtrlC:
+					// TODO: needs locking?
+					t.cancel = true
+					break loop
+				}
+			}
+		}
+	}()
+
+	updateAndEvaluate := func() (bool, error) {
+		var m daemon.Status
+
+		err := jsonRequestResponse(httpc, daemon.ControlJobEndpointStatus,
+			struct{}{},
+			&m,
+		)
+		if err != nil {
+			return true, fmt.Errorf("can't get status from daemon: %w", err)
+		}
+
+		t.report = m.Jobs
+
+		return t.evaluate()
+	}
+	updateAndEvaluate()
+
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
+ticker:
+	for range ticker.C {
+		if t.cancel {
+			break ticker
+		}
+
+		done, err := updateAndEvaluate()
+		if err != nil {
+			return err
+		}
+		if done {
+			break ticker
+		}
+	}
+
+	return nil
+}
+
+func (t *wui) evaluate() (bool, error) {
+	v, ok := t.report[t.jobName]
+	if !ok {
+		return true, fmt.Errorf("job not in status: %v", t.jobName)
+	}
+
+	if v.Type == job.TypePush || v.Type == job.TypePull {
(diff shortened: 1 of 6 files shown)


client/wait.go
```diff
@@ -0,0 +1,248 @@
...
+type wui struct {
+	//lock      sync.Mutex //For report and error
+	report map[string]*job.Status
+
+	jobName string
+}
+
+var WaitCmd = &cli.Subcommand{
+	Use:   "wait JOB",
+	Short: "block and wait until the specified job is done or has failed",
```
problame, line 28: need to explain what `done` means somehwere

client/wait.go
```diff
@@ -0,0 +1,248 @@
...
+	Run: func(ctx context.Context, subcommand *cli.Subcommand, args []string) error {
+		return runWaitCmd(subcommand.Config(), args)
+	},
+}
+
+// adapted from status.go
```
problame, line 34: perhaps we should re-use the `client` from the status-v2 work https://github.com/zrepl/zrepl/pull/297

client/wait.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/zrepl/zrepl/cli"
+	"github.com/zrepl/zrepl/config"
+	"github.com/zrepl/zrepl/daemon"
+	"github.com/zrepl/zrepl/daemon/job"
+	"github.com/zrepl/zrepl/daemon/pruner"
+	"github.com/zrepl/zrepl/daemon/snapper"
+	"github.com/zrepl/zrepl/replication/report"
+)
+
+type wui struct {
```
problame, line 19: confusing type name

client/wait.go
```diff
@@ -0,0 +1,248 @@
...
+
+		done := replicationDone && senderPruningDone && receiverPruningDone
+
+		if v.Type == job.TypePush {
+			snapperDone, err := evaluateSnapperReport(activeStatus.Snapshotting)
+			if err != nil {
+				return true, err
+			}
+
+			done = done && snapperDone
```
problame, line 120: Handling the snapshotting is tricky: snapshotting is never exactly 'done'. Maybe the command should be more like `zrepl wait-job JOBNAME --replication [true,false] --snapshotting [true,false] --pruning [true,false]` ?

See first commit in status-v2 `add signal 'snapshot', rename existing signal 'wakeup' to 'replication'` [link](https://github.com/zrepl/zrepl/pull/297/commits/a1a0c9f30552dbe2d42fd19a88e4dd40f88d8b57)

client/wait.go
```diff
@@ -0,0 +1,248 @@
...
+	if r.Error != "" {
+		return true, fmt.Errorf("pruning error: %v", r.Error)
+	}
+
+	prunerState, err := pruner.StateString(r.State)
+	if err != nil {
+		return true, fmt.Errorf("parsing pruning state %v: %w", r.State, err)
+	}
+
+	switch prunerState {
```
problame, line 176: we should have a `prunerState.IsTerminal()` getter, use it both here and in zrepl status

problame (commented):
Thanks for putting this together.

I think this work would benefit from queuing up behind [status-v2](https://github.com/zrepl/zrepl/pull/297/commits), in particular the first commit that splits up signal `wakeup` into signals `replication` and signal `snapshot`.

riyad:
Now that there's a new RC with the status-v2 changes, is it safe to start with a reimplementation of this feature?

client/wait.go
riyad: I'll look into it :slightly_smiling_face:

client/wait.go
riyad: yeah :sweat_smile:  ... did I mention it was adapted from the status code :wink:

client/wait.go
riyad: hmmm ... I was wondering if there're situations where a job finished fast enough that the wait after a signal would be stuck waiting on a job thinking that it's not started yet.

client/wait_job.go
```diff
@@ -0,0 +1,229 @@
...
+			return true, err
+		}
+
+		snapperDone, err := evaluateSnapperStatus(snapStatus.Snapshotting)
+		if err != nil {
+			return true, err
+		}
+
+		return pruningDone && snapperDone, nil
+	} else if j.Type == job.TypeSource /*|| j.Type == job.TypeSink*/ {
```
riyad, line 130: I'm not sure what to do here. The status code doesn't even mention `job.TypeSink`.

client/wait.go
riyad: I'd suggest keeping it simple at first and see if there's actually a need for sub-job granularity.

problame:
> Now that there's a new RC with the status-v2 changes, is it safe to start with a reimplementation of this feature?

Sorry for not getting back to this sooner. @InsanePrawn and myself started drafting a new architecture for 'signal', 'reset', **and 'wakeup'** yesterday. The gist of the idea is to generalize the `invocation` id that we keep in `active.go` and use the tuple `(job, invocation)` to wait for a job to finish. This architecture will also allow for race-free `zrepl signal --wait JOB [replication|snapshotting]`, which we think is a pretty neat addition. Waiting for job completion is still poll based, but won't have to rely on polling the `status`.
Thus, I'd recommend waiting until next week - by then we'll hopefully have a draft PR out so we can see whether the design works out in practice.

Again, sorry for not getting back to this sooner :/

riyad:
Awesome! tThis was a step up from shell magic and your solution is a huge step up from that. I'll take race-free any time. :+1:
This task was still a nice mental jog for me. :smile:

PR #442, opened 2021-03-15
client/status: allow raw mode without a tty

cole-h:
A fairly common pattern when presented with JSON is to pipe it to `jq`.

This also allows one to `zrepl status --mode raw > log` and operate on
the JSON there.

---

Prior to this change, attempting to pipe to `jq` would greet me with:

```
$ doas zrepl status --mode raw | jq
error: stdout is not a tty, please use --mode dump
```

Now:

```
$ doas zrepl status --mode raw | jq | head
{
  "Jobs": {
    "_control": {
      "internal": null,
      "type": "internal"
    },
    "bpool_sink": {
      "sink": {
        "Snapper": null
      },
```

### DIFF
--- a/client/status/status.go
+++ b/client/status/status.go
@@ -71,12 +71,16 @@ func runStatusV2Command(ctx context.Context, config *config.Config, args []strin
 
 	mode := statusv2Flags.Mode.Value().(statusv2Mode)
 
-	if !isatty.IsTerminal(os.Stdout.Fd()) && mode != StatusV2ModeDump {
-		usemode, err := statusv2Flags.Mode.InputForChoice(StatusV2ModeDump)
+	if !isatty.IsTerminal(os.Stdout.Fd()) && mode != StatusV2ModeDump && mode != StatusV2ModeRaw {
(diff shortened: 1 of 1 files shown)


client/status/status.go
```diff
@@ -71,7 +71,7 @@ func runStatusV2Command(ctx context.Context, config *config.Config, args []strin
 
 	mode := statusv2Flags.Mode.Value().(statusv2Mode)
 
-	if !isatty.IsTerminal(os.Stdout.Fd()) && mode != StatusV2ModeDump {
+	if !isatty.IsTerminal(os.Stdout.Fd()) && mode != StatusV2ModeDump && mode != StatusV2ModeRaw {
```
grahamc, line 74: I think the error message should also suggest the json mode?

cole-h: Thanks for noticing; resolved in force-push.

InsanePrawn (approved):
LGTM

problame (approved):
Yup, sorry I let this slip through. Should we issue another rc immediately or wait a few days?

cole-h:
No worries!

I think it's fine to wait a few days in order to potentially let some more fixes get in.

PR #445, opened 2021-03-16
systemd: enable restarting zrepl service

MAFLO321:
This is useful if the zrepl daemon has an error for any reason and exits.

### DIFF
--- a/dist/systemd/zrepl.service
+++ b/dist/systemd/zrepl.service
@@ -4,6 +4,7 @@ Documentation=https://zrepl.github.io
 
 [Service]
 Type=simple
+Restart=always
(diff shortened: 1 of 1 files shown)


InsanePrawn:
What about cooldowns and limits?
I wouldn't want to be running the equivalent of `while true; do zrepl daemon; done` while zrepl's crashing. :)

problame:
I'm quite hesitant to put `Restart=always` into a systemd unit. The zrepl daemon should never be crashing. If it does it's a bug that needs to be reported and fixed. Does `Restart=always`

* keep the core dumps and/or stack trace around somewhere?
* notify the user / expose some metric that can be monitored easily?

problame:
Closing this due to reasons [outlined above](https://github.com/zrepl/zrepl/pull/445#issuecomment-804447283).

PR #452, opened 2021-03-23
0.4rc1: Revert to zrepl 0.3 style signalling

InsanePrawn:
Revert the split signalling merged as part of status-v2.

As mentioned in https://github.com/zrepl/zrepl/pull/436#issuecomment-804445134, we're drafting an overhaul of the signalling mechanisms and better decoupling for jobs' tasks, but this will not land in 0.4.

Revert 0.4 (currently RC1) to use the old unified signalling (`zrepl signal wakeup $job` and `zrepl signal reset $job` being the only options) for now.


docs/changelog.rst
```diff
@@ -32,7 +32,6 @@ We use the following annotations for classifying changes:
 0.4.0
 -----
 
-* |break| Change syntax to trigger a job replication, rename ``zrepl signal wakeup JOB`` to ``zrepl signal replication JOB``
```
InsanePrawn, line 35: @problame not sure if dropping this line is the way to go. alternatively we could do a strikethrough and add something like 'postponed to a future release'
We definitely need to mention this revert in the RC2 release notes tho.

PR #454, opened 2021-03-26
Change suggested metrics port

### LINKED ISSUE #444: prometheus exporter: use better default port number & maybe consider registering it
Current documentation suggests using port `9091`, but that is the port that Prometheus server uses, I don't think that a good default to suggest for a single export client.

You could register a "default port" [in this page](https://github.com/prometheus/prometheus/wiki/Default-port-allocations), so that all `zrepl` installs can have a standard value to use.

lapo-luchini:
Closes #444.

### DIFF
--- a/config/config_global_test.go
+++ b/config/config_global_test.go
@@ -70,9 +70,9 @@ func TestPrometheusMonitoring(t *testing.T) {
 global:
   monitoring:
     - type: prometheus
-      listen: ':9091'
+      listen: ':9811'
 `)
-	assert.Equal(t, ":9091", conf.Global.Monitoring[0].Ret.(*PrometheusMonitoring).Listen)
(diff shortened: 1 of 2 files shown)


InsanePrawn (approved):
Cool, thanks for taking care of registering the port with the prometheus project as well!

Not sure if we also want a changelog entry for this doc change.

InsanePrawn:
I found a couple of additional references to the old port and decided it was quickest to just fix them, I've taken the liberty of pushing my additions with an enhanced commit message to your branch.

problame:
Thanks for getting this done!

problame:
committed in https://github.com/zrepl/zrepl/commit/3b5a1a8b9a14c167d1af3ab7af115d5f9ffe9c2f

Issue #403, opened 2020-12-22
zrepl replicates every snapshot including those created outside zrepl


calistoc:
Hello,
just to ask some help please as I was searching the documentation but I’m confused and not sure whether I’m missing something.

For what I see zrepl replicates all the snapshots in a filesystem including those snaps created outside zrepl, for example I have some snaps created to perform a daily backup that I would like not to be considered by zrepl but they get replicated once I wakeup the jobs.
This happens for push jobs with periodic snapshotting and also with jobs with manual snapshotting where I need to wakeup the jobs via crontab and where I cannot specify a prefix in the snapshotting configuration:

snapshotting:
    type: manual
    prefix: zrepl_ (it gives this error:  field prefix not found in type config.SnapshottingManual)

It seems to me that the snapshotting prefix for push jobs with periodic snapshotting is only used for pruning but it is not used to select the snapshots to be replicated, is this correct?
Is there a way to have zrepl replicating only its own snapshots or better only those specified by a prefix?

Thank you

problame:
Hi, thanks for describing your use case.

> It seems to me that the snapshotting prefix for push jobs with periodic snapshotting is only used for pruning but it is not used to select the snapshots to be replicated, is this correct?

Actually, it's even less than that. The prefix you specify in the `snapshotting` section is really only used to define the snapshot name, nothing more, nothing less.

>  Is there a way to have zrepl replicating only its own snapshots or better only those specified by a prefix?

No, currently there isn't a way to tell zrepl to do that.

- #248 requests prefix-based filtering of snapshots for replication. It seems that if we implemented prefix-based filtering, your problem would be solved. Wouldn't it?
- It seems like you assumed zrepl would somehow 'own' the snapshots it created and either only act on the snapshots it created or treat them specially during replication or pruning. That is currently not the case - replication and pruning do not know whether a snapshot was created by zrepl or something else. But it's a very widespread belief, and zrepl should probably develop into a direction where snapshot management and replication are more tightly integrated.

If you feel like #248 would be the idea solution, please subscribe there and post a proposal for the configuration syntax in your scenario.

calistoc:
Thanks for your feedbacks

Yes I think the [#248](https://github.com/zrepl/zrepl/issues/248) request would solve my problem.
About the proposal for the configuration syntax, I would add the following optional parameter in the configuration job, either push, pull or source, something like this:

- type: push
  name: to_server2
  replication_snapshots_prefix: zrepl_
…

[...]

PR #459, opened 2021-04-07
job snapshots prefix for replication filtering

calistoc:
This PR will allow job snapshots prefix for replication filtering.

Added optional parameter replication_snapshots_prefix that will tell zrepl to replicate only snapshots with the prefix provided, it is configurable only for push and source jobs and will be filtering the snapshots from the sender side. If not used zrepl will keep working as it used to replicating every snapshot created in the filesystem.

zrepl.yml syntax example:

-  type: push
   replication_snapshots_prefix: “zrepl_”
   …

Should address https://github.com/zrepl/zrepl/issues/403

Please give it a try and see if it helps

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -123,14 +123,16 @@ type PropertyRecvOptions struct {
 }
 
 type PushJob struct {
-	ActiveJob    `yaml:",inline"`
-	Snapshotting SnapshottingEnum  `yaml:"snapshotting"`
-	Filesystems  FilesystemsFilter `yaml:"filesystems"`
-	Send         *SendOptions      `yaml:"send,fromdefaults,optional"`
+	ActiveJob     `yaml:",inline"`
+	Snapshotting  SnapshottingEnum  `yaml:"snapshotting"`
+	Filesystems   FilesystemsFilter `yaml:"filesystems"`
+	Send          *SendOptions      `yaml:"send,fromdefaults,optional"`
+	JobSnapPrefix string            `yaml:"replication_snapshots_prefix,optional"`
 }
 
 func (j *PushJob) GetFilesystems() FilesystemsFilter { return j.Filesystems }
 func (j *PushJob) GetSendOptions() *SendOptions      { return j.Send }
+func (j *PushJob) GetJobSnapPrefix() string          { return j.JobSnapPrefix }
 
 type PullJob struct {
 	ActiveJob `yaml:",inline"`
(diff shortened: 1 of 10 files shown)


zfs/versions.go
```diff
@@ -211,6 +211,13 @@ func (o *ListFilesystemVersionsOptions) typesFlagArgs() string {
 }
 
 func (o *ListFilesystemVersionsOptions) matches(v FilesystemVersion) bool {
+	// Workaround when using JobSnapPrefix because bookmarks are not following
+	// the snapshotting prefix but they have fixed prefix:
+	// replicationCursorBookmarkNamePrefix = "zrepl_CURSOR"
+	if len(o.ShortnamePrefix) > 0 && len(o.Types) == 0 && v.Type == Bookmark && strings.HasPrefix(v.Name, "zrepl_CURSOR") {
```
problame, line 217: We should just allow all bookmarks, probably rename `ShortnamePrefix` to `SnapPrefix`.

config/config.go
```diff
@@ -123,14 +123,16 @@ type PropertyRecvOptions struct {
...
 type PushJob struct {
-	ActiveJob    `yaml:",inline"`
-	Snapshotting SnapshottingEnum  `yaml:"snapshotting"`
-	Filesystems  FilesystemsFilter `yaml:"filesystems"`
-	Send         *SendOptions      `yaml:"send,fromdefaults,optional"`
+	ActiveJob     `yaml:",inline"`
+	Snapshotting  SnapshottingEnum  `yaml:"snapshotting"`
+	Filesystems   FilesystemsFilter `yaml:"filesystems"`
+	Send          *SendOptions      `yaml:"send,fromdefaults,optional"`
+	JobSnapPrefix string            `yaml:"replication_snapshots_prefix,optional"`
```
problame, line 130: Needs coverage in docs.

endpoint/endpoint.go
```diff
@@ -116,7 +117,11 @@ func (s *Sender) ListFilesystemVersions(ctx context.Context, r *pdu.ListFilesyst
 	if err != nil {
 		return nil, err
 	}
-	fsvs, err := zfs.ZFSListFilesystemVersions(ctx, lp, zfs.ListFilesystemVersionsOptions{})
+
+	fsvs, err := zfs.ZFSListFilesystemVersions(ctx, lp, zfs.ListFilesystemVersionsOptions{
+		ShortnamePrefix: s.config.JobSnapPrefix,
```
problame, line 122: Maybe we should have something more flexible than just prefix matching?
See my ideas in https://github.com/zrepl/zrepl/issues/403#issuecomment-820698946

problame (commented):
Captured some of my comments from the thread in #403 so I don't forget them if we chose to go with the approach implemented in this PR.

zfs/versions.go
calistoc: done, thanks

config/config.go
calistoc: added some docs, I hope I did it right and also it is enough, feel free to change anything.

endpoint/endpoint.go
calistoc: Sure there is space for improvements, I’ll think about it but I can’t promise anything in a short time.

config/config.go
calistoc: I got some errors with the circleci checks.
It looks better now but the test-go-on-latest-go-release check is still failing, no idea why yet, any help is welcome:)

wrenix:
any update on this?

PR #461, opened 2021-04-08
Documentation: debian/ubuntu-snippet with sudo

mek-apelsin:
I generally like when snippets are provided in a way which could be used without running as root, and uses sudo when applicable. This change allows for this. 

It will, however print out one extra line, which is possible to remove by adding '>/dev/null' after '/etc/apt/sources.list.d/zrepl.list'.

### DIFF
--- a/docs/installation/apt-repos.rst
+++ b/docs/installation/apt-repos.rst
@@ -13,13 +13,13 @@ The following snippet configure the repository for your Debian or Ubuntu release
 
 ::
 
-    apt update && apt install curl gnupg lsb-release; \
+    sudo apt update && sudo apt install curl gnupg lsb-release; \
     ARCH="$(dpkg --print-architecture)"; \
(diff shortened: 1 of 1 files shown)


problame:
Thanks for this contribution.

> It will, however print out one extra line, which is possible to remove by adding '>/dev/null' after '/etc/apt/sources.list.d/zrepl.list'.

Why is that?

mek-apelsin:
> Thanks for this contribution.
> 
> > It will, however print out one extra line, which is possible to remove by adding '>/dev/null' after '/etc/apt/sources.list.d/zrepl.list'.
> 
> Why is that?

`sudo tee $file` will run tee as root, write the result to $file and spit out the result to standard out. 

I usually run sudo tee >/dev/null $file to make it more obvious, but shfmt (and maybe shellcheck) frowns on this.

PR #466, opened 2021-04-12
Use ISO-8601 timestamp for snapshots

### LINKED ISSUE #465: The timestamp format is non-standard, but could be
Using the current timestamp format incurs a cost for everybody, since it isn't a standard and trivially readable format. Might I propose changing the format to something more standard? For example, ISO8601 supports:

* `%Y-%m-%dT%H:%M:%SZ`
*  `%Y-%m-%dT%H%M%SZ`

both of which should be no trouble for ZFS.

grahamc:
Close #465

### DIFF
--- a/daemon/snapper/snapper.go
+++ b/daemon/snapper/snapper.go
@@ -250,7 +250,7 @@ func snapshot(a args, u updater) state {
 	anyFsHadErr := false
 	// TODO channel programs -> allow a little jitter?
 	for fs, progress := range plan {
-		suffix := time.Now().In(time.UTC).Format("20060102_150405_000")
+		suffix := time.Now().In(time.UTC).Format("2006-01-02T150405.000Z")
 		snapname := fmt.Sprintf("%s%s", a.prefix, suffix)
 
 		ctx := logging.WithInjectedField(a.ctx, "fs", fs.ToString())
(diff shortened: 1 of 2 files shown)


grahamc:
I am not sure that the failing test is due to this PR...

grahamc:
I just fixup'd the format to use a colon free format thanks to @debliski's point in the issue.

problame:
Let's wait a little longer and see whether there's any more discussion in #465

problame:
Regardless, this change should also be documented in `changelog.rst`

problame:
Sorry for taking so long to get back to this. 

So, apart from

- [ ] `changelog.rst` note

the problem is that there might be [Samba server configs](https://www.samba.org/samba/docs/current/man-html/vfs_shadow_copy2.8.html) out there that rely on the current date format for the shadow copies feature (search for `shadow:format = format specification for snapshot names` on that site).
As far as I can tell, you can't have multiple formats simultaneously => those users are stuck with the old format, or lose access to old snapshots.
How do we address this?

problame:
@grahamc are you interested in turning this into a config field that defaults to the old format, but allows users to specify an arbitrary format? Or, maybe choose from a set of pre-defined formats?
Thinking something like

```
...
- ...
  snapshotting:
    type: periodic
    prefix: zrepl_
    utc_timestamp_format: "dense" | "iso-8601"
    interval: 10m
    
```

problame:
Otherwise, please close the PR because I don't want to break existing Samba configs, etc. ([See above](https://github.com/zrepl/zrepl/pull/466#issuecomment-876568393)]

problame:
Closing due to inactivity.

Issue #408, opened 2021-01-03
automatic conflict resolution


problame:
zrepl should be able to resolve replication conflicts automatically.

## Background / Use Cases

**Replication conflicts** are situations where there is *something* on the receiving side that prevents replication of one or more file systems from the sender to the receiver.
Examples and use cases:

* receiver-only snapshots
  * `zfs snapshot -r receivingpool@pre_upgrade_snapshot`
  * forgotten exception in auto-snaphotting tool on receive side
  * ...
* unsnapshotted modifications on the receiver => recv fails with characteristic error;  cc #405
* renamed and/or re-created file systems on the sender => looks like complete divergence to replication algorithm; cc #70 

## UX

* zrepl should continue to opt for non-destructive-by-default behavior
* Renames should be detected automatically, e.g. by tracking GUIDs that are invariant to replication cc #70.
  Renames are thus not actually part of the conflict resolution UX.
* With the daemon, no user interaction is required  for conflict resolution.
* Instead, a conflict resolution policy defined in the config is applied automatically to all filesystems matched by a job.
* If the conflict resolution gives up, the user is informed via logs, metrics and zrepl status.
* **Optional** The `zrepl signal` (or similar) command allows the user to override the conflict resolution policy on a per-filesystem basis for the next replication attempt.

Config Syntax:

```yaml
type: push # active side
...
conflict_resolution:

  # what to do if the receiver diverged and we can't replicate:
  receiver_diverged:

    # if there are (only) local modifications but no snapshots:
    unsnapshotted_modifications:
      # Policy Option #1: fail filesystems affected by this
      policy:
        type: fail
      # Policy Option #2: rollback filesystem to latest snapshot
      policy:
        type: rollback_to_latest_snapshot
        max_written: 10MiB # max value for the `written` zfs property of the dataset (fail otherwise)

    # if there are snapshots on the receiver that prevent replication
    receiver_only_snapshots:
      # Policy Option #1: fail filesystems affected by this
      policy:
        type: fail

      # Policy Option #2: destroy the receiver-only snapshots that prevent replication
      policy:
        type: destroy
        max_time: 1d # the max time window that is allowed to be rolled
        max_count: 10 # the max number of snapshots that are allowed to be rolled back. -1 for inifinity, 0 is invalid
        snapshots: # regex that match the snapshot names (NOTE: syntax should be the same as for snapshot filtering)
          - regex: "^zrepl_.*"
            action: rollback
          - regex: ".*"
            action: prohibit

[...]

PR #470, opened 2021-04-15
introduce conflict_resolution.initial_replication.replicate_all_snapshots

### LINKED ISSUE #187: Only last snapshot is backeded up in 'manual' mode
When the backup mechanism encounters a filesystem that has not been created yet on the backup side, only the last snapshot is transferred.

### Steps to reproduce

**Preparation**

[YAML config file](https://github.com/zrepl/zrepl/files/3293623/zrepl.yml.txt)

```
# zfs list -o name -r -t filesystem,snapshot
NAME
DestinationPool
SourcePool
SourcePool/MyFilesystem
SourcePool/MyFilesystem@1
SourcePool/MyFilesystem@2
SourcePool/MyFilesystem@3
SourcePool/MyFilesystem@4
# systemctl start zrepl
# zrepl signal wakeup Backup
# zfs list -o name -r -t filesystem,snapshot
```
**Expected Result**
```
NAME
DestinationPool
DestinationPool/MyBackup
DestinationPool/MyBackup/SourcePool
DestinationPool/MyBackup/SourcePool/MyFilesystem
DestinationPool/MyBackup/SourcePool/MyFilesystem@1
DestinationPool/MyBackup/SourcePool/MyFilesystem@2
DestinationPool/MyBackup/SourcePool/MyFilesystem@3
DestinationPool/MyBackup/SourcePool/MyFilesystem@4
SourcePool
SourcePool/MyFilesystem
SourcePool/MyFilesystem@1
SourcePool/MyFilesystem@2
SourcePool/MyFilesystem@3
SourcePool/MyFilesystem@4
```
**Result**
```
NAME
DestinationPool
DestinationPool/MyBackup
DestinationPool/MyBackup/SourcePool
DestinationPool/MyBackup/SourcePool/MyFilesystem
DestinationPool/MyBackup/SourcePool/MyFilesystem@4
SourcePool
SourcePool/MyFilesystem
SourcePool/MyFilesystem@1
SourcePool/MyFilesystem@2
SourcePool/MyFilesystem@3
SourcePool/MyFilesystem@4
```

grahamc:
By configuring your push or pull job with:

```yaml
conflict_resolution:
  initial_replication:
    replicate_all_snapshots: true
```

when a dataset is transferred for the first time, all snapshots on
that dataset will be transferred. This is disabled by default, and
only the most recent snapshot will be transferred.

Based on the names in https://github.com/zrepl/zrepl/issues/408#issuecomment-795329846
though the more complicated options suggested in that issue are
more Go than I know.

Closes #187

### DIFF
--- a/docs/configuration/conflict_resolution.rst
+++ b/docs/configuration/conflict_resolution.rst
@@ -1,0 +1,27 @@
+.. include:: ../global.rst.inc
+
+
+Conflict Resolution Options
+===========================
+
+
+::
+
+   jobs:
+   - type: push
+     filesystems: ...
+     conflict_resolution:
+       initial_replication:
+           send_all_snapshots: false # false,true
+
+     ...
+
+.. _conflict_resolution-initial_replication-option-send_all_snapshots:
+
+
+``send_all_snapshots`` option
+-----------------------------
+
+The ``send_all_snapshots`` options control the initial transfer of push and pull jobs.
+Left ``false`` (the default), the initial transfer will only transfer the most recent snapshot.
+If ``send_all_snapshots`` is ``true`` for the very first transfer of a dataset, all historical snapshots will be transfered.
--- a/replication/logic/replication_logic.go
+++ b/replication/logic/replication_logic.go
@@ -234,25 +234,38 @@ func NewPlanner(secsPerState *prometheus.HistogramVec, bytesReplicated *promethe
 		promBytesReplicated: bytesReplicated,
 	}
 }
-func resolveConflict(conflict error) (path []*pdu.FilesystemVersion, msg string) {
+func resolveConflict(conflict error, policy PlannerPolicy) (initialSend bool, path []*pdu.FilesystemVersion, msg string) {
 	if noCommonAncestor, ok := conflict.(*ConflictNoCommonAncestor); ok {
 		if len(noCommonAncestor.SortedReceiverVersions) == 0 {
-			// TODO this is hard-coded replication policy: most recent snapshot as source
+			// NOTE: I did not keep this in sync with listStaleFiltering, it isn't clear to me what the right thing to do is.
 			// NOTE: Keep in sync with listStaleFiltering, it depends on this hard-coded assumption
-			var mostRecentSnap *pdu.FilesystemVersion
-			for n := len(noCommonAncestor.SortedSenderVersions) - 1; n >= 0; n-- {
-				if noCommonAncestor.SortedSenderVersions[n].Type == pdu.FilesystemVersion_Snapshot {
-					mostRecentSnap = noCommonAncestor.SortedSenderVersions[n]
-					break
+			if policy.InitiallyAllSnapshots {
+				var allSnapshots []*pdu.FilesystemVersion
+				for n := 0; n < len(noCommonAncestor.SortedSenderVersions); n++ {
+					if noCommonAncestor.SortedSenderVersions[n].Type == pdu.FilesystemVersion_Snapshot {
+						allSnapshots = append(allSnapshots, noCommonAncestor.SortedSenderVersions[n])
+					}
 				}
+				if len(allSnapshots) == 0 {
+					return true, nil, "no snapshots available on sender side"
+				}
+				return true, allSnapshots, "start replication at the oldest snapshot"
+			} else {
+				var mostRecentSnap *pdu.FilesystemVersion
+				for n := len(noCommonAncestor.SortedSenderVersions) - 1; n >= 0; n-- {
+					if noCommonAncestor.SortedSenderVersions[n].Type == pdu.FilesystemVersion_Snapshot {
+						mostRecentSnap = noCommonAncestor.SortedSenderVersions[n]
+						break
+					}
+				}
+				if mostRecentSnap == nil {
+					return true, nil, "no snapshots available on sender side"
+				}
+				return true, []*pdu.FilesystemVersion{mostRecentSnap}, fmt.Sprintf("start replication at most recent snapshot %s", mostRecentSnap.RelName())
 			}
-			if mostRecentSnap == nil {
-				return nil, "no snapshots available on sender side"
-			}
-			return []*pdu.FilesystemVersion{mostRecentSnap}, fmt.Sprintf("start replication at most recent snapshot %s", mostRecentSnap.RelName())
 		}
 	}
-	return nil, "no automated way to handle conflict type"
+	return false, nil, "no automated way to handle conflict type"
 }
 
 func (p *Planner) doPlanning(ctx context.Context) ([]*Filesystem, error) {
@@ -453,10 +466,11 @@ func (fs *Filesystem) doPlanning(ctx context.Context) ([]*Step, error) {
 			})
 		}
 	} else { // resumeToken == nil
+		var initialSend = false
 		path, conflict := IncrementalPath(rfsvs, sfsvs)
 		if conflict != nil {
 			var msg string
-			path, msg = resolveConflict(conflict) // no shadowing allowed!
+			initialSend, path, msg = resolveConflict(conflict, fs.policy) // no shadowing allowed!
 			if path != nil {
 				log(ctx).WithField("conflict", conflict).Info("conflict")
 				log(ctx).WithField("resolution", msg).Info("automatically resolved")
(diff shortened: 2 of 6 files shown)


grahamc:
Note that I don't really know Go.

replication/logic/replication_logic.go
```diff
@@ -234,25 +234,38 @@ func NewPlanner(secsPerState *prometheus.HistogramVec, bytesReplicated *promethe
...
+				for n := 0; n < len(noCommonAncestor.SortedSenderVersions); n++ {
+					if noCommonAncestor.SortedSenderVersions[n].Type == pdu.FilesystemVersion_Snapshot {
+						allSnapshots = append(allSnapshots, noCommonAncestor.SortedSenderVersions[n])
+					}
 				}
+				if len(allSnapshots) == 0 {
+					return true, nil, "no snapshots available on sender side"
+				}
+				return true, allSnapshots, "start replication at the oldest snapshot"
+			} else {
```
kalbasit, line 253: Since you're returning in the if branch, you don't need the else branch.

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow (applies to non-error handling code as well)

replication/logic/replication_logic.go
```diff
@@ -234,25 +234,38 @@ func NewPlanner(secsPerState *prometheus.HistogramVec, bytesReplicated *promethe
...
-				if noCommonAncestor.SortedSenderVersions[n].Type == pdu.FilesystemVersion_Snapshot {
-					mostRecentSnap = noCommonAncestor.SortedSenderVersions[n]
-					break
+			if policy.InitiallyAllSnapshots {
+				var allSnapshots []*pdu.FilesystemVersion
+				for n := 0; n < len(noCommonAncestor.SortedSenderVersions); n++ {
+					if noCommonAncestor.SortedSenderVersions[n].Type == pdu.FilesystemVersion_Snapshot {
+						allSnapshots = append(allSnapshots, noCommonAncestor.SortedSenderVersions[n])
+					}
 				}
```
kalbasit, line 248:
```suggestion
				for _, ver := range noCommonAncestor.SortedSenderVersions {
					if ver.Type == pdu.FilesystemVersion_Snapshot {
						allSnapshots = append(allSnapshots, ver)
					}
				}
```

replication/logic/replication_logic.go
```diff
@@ -453,10 +466,11 @@ func (fs *Filesystem) doPlanning(ctx context.Context) ([]*Step, error) {
 			})
 		}
 	} else { // resumeToken == nil
+		var initialSend = false
```
kalbasit, line 469:
```suggestion
		var initialSend bool
```

All Go types defaults to their zero-value for the type. false for bool, 0 for numbers (int, float, etc..), empty string for strings, nil for pointers etc...

replication/logic/replication_logic.go
```diff
@@ -469,29 +483,23 @@ func (fs *Filesystem) doPlanning(ctx context.Context) ([]*Step, error) {
...
 
+		if initialSend {
+			// Add a nil path to the beginning so our loop over each step starts with a nil...FIRSTSNAPSHOT step.
+			// The nil `from` is what indicates the send is not incremental.
+			path = append([]*pdu.FilesystemVersion{nil}, path...)
+		}
+
 		steps = make([]*Step, 0, len(path)) // shadow
-		if len(path) == 1 {
+		for i := 0; i < len(path)-1; i++ {
```
kalbasit, line 493: This could also be a range. Simpler and optimized for slices.

replication/logic/replication_logic.go
```diff
@@ -469,29 +483,23 @@ func (fs *Filesystem) doPlanning(ctx context.Context) ([]*Step, error) {
 			return nil, conflict
 		}
 
+		if initialSend {
+			// Add a nil path to the beginning so our loop over each step starts with a nil...FIRSTSNAPSHOT step.
+			// The nil `from` is what indicates the send is not incremental.
+			path = append([]*pdu.FilesystemVersion{nil}, path...)
```
kalbasit, line 489: I'm not a big fan of this as it creates a hidden assumption in your slice. Where is the code that expects this nil?

kalbasit (changes requested):

replication/logic/replication_logic.go
grahamc: The code used to look like this in the "initial send" case (used to be called `len(..) == 1`):

```

			steps = append(steps, &Step{
				parent:   fs,
				sender:   fs.sender,

				from:    nil,
				to:      path[0],
				encrypt: fs.policy.EncryptedSend,
			})
```

seen in the diff below.

replication/logic/replication_logic.go
grahamc: I think I need the range in this case to get i, since I'm essentially windowing across the slice.

kalbasit: Oh right, sorry I missed `path[i+1]`.

replication/logic/replication_logic.go
kalbasit: I see. It feels hacky tbh. Personally, I'd go for this:

```
steps = make([]*Step, 0, len(path)) // shadow
if initialSend {
	steps = append(steps, &Step{
			parent:   fs,
			sender:   fs.sender,
			receiver: fs.receiver,

			from:    nil,
			to:      path[0],
			encrypt: fs.policy.EncryptedSend,
	})
}

for i := 0; i < len(path)-1; i++ {
	steps = append(steps, &Step{
		parent:   fs,
		sender:   fs.sender,
		receiver: fs.receiver,

		from:    path[i],
		to:      path[i+1],
		encrypt: fs.policy.EncryptedSend,
	})
}
```

or with something like this:

```
steps = make([]*Step, 0, len(path)) // shadow

appendStep := func(a, b *Step) {
	steps = append(steps, &Step{
		parent:   fs,
		sender:   fs.sender,
		receiver: fs.receiver,

		from:    a,
		to:      b,
		encrypt: fs.policy.EncryptedSend,
	})
}

if initialSend {
	appendStep(nil, path[0])
}

for i := 0; i < len(path)-1; i++ {
	appendStep(path[i], path[i+1])
}
```

replication/logic/replication_logic.go
```diff
@@ -234,25 +234,38 @@ func NewPlanner(secsPerState *prometheus.HistogramVec, bytesReplicated *promethe
 		promBytesReplicated: bytesReplicated,
 	}
 }
-func resolveConflict(conflict error) (path []*pdu.FilesystemVersion, msg string) {
+func resolveConflict(conflict error, policy PlannerPolicy) (initialSend bool, path []*pdu.FilesystemVersion, msg string) {
 	if noCommonAncestor, ok := conflict.(*ConflictNoCommonAncestor); ok {
 		if len(noCommonAncestor.SortedReceiverVersions) == 0 {
-			// TODO this is hard-coded replication policy: most recent snapshot as source
+			// NOTE: I did not keep this in sync with listStaleFiltering, it isn't clear to me what the right thing to do is.
```
problame, line 240: This comment is referring to the following comment in `listStaleFiltering`

https://github.com/zrepl/zrepl/blob/1c937e58f7435c509ba97baab4cfdd9958d843ae/endpoint/endpoint_zfs_abstraction.go#L824-L828

replication/logic/replication_logic.go
problame: This is policy code, not error handling. I prefer policy code to be a decision tree <=> if + else

replication/logic/replication_logic.go
problame: I like @kalbasit 's second suggestion

docs/configuration/conflict_resolution.rst
```diff
@@ -0,0 +1,27 @@
...
+
+
+::
+
+   jobs:
+   - type: push
+     filesystems: ...
+     conflict_resolution:
+       initial_replication:
+           send_all_snapshots: false # false,true
```
problame, line 15: We need to iterate on this, since, as suggested in https://github.com/zrepl/zrepl/issues/408#issuecomment-795329846, there are three possible options:

* fail
* most recent
* all

I think we should add the `fail` policy option to this PR as well. The benefit will be that we're forced to define a config syntax that deals with non-boolean policy options, leading to less breaking changes down the road.
I'm not convinced that the `policy: { type: ... }` schema in #408 is a good syntax (too verbose), feel free to come up with a better one. (The advantage of `type: ` is that our config parser has infrastructure to handle this kind of 'enum').

grahamc: Yeah, I wanted to go this route but I didn't quite understand how to do it. I ran out of time to pursue this further right now, though. I'm not sure but I think there might also be a bug in the scenario where the first replication of many snapshots gets interrupted and isn't resumed properly. Sorry for dropping it here, I could close it if you would like.

problame:
It seems like this stalled due to https://github.com/zrepl/zrepl/pull/470#discussion_r614653598

@grahamc do you have resources to pick this up? 

I think what we want is:

```yaml
     filesystems: ...
     conflict_resolution:
       initial_replication: send_all | send_most_recent | fail
```

problame:
I am finalizing this work in branch `initial-replication-policy`

PR #472, opened 2021-04-19
Fix typo in example derive -> drive

splaspood:
Just fixing a small typo in the documentation comments at the top of the file.   'drive' is instead 'derive'


problame:
Thanks!

PR #481, opened 2021-04-30
add per filesystem replication error to prometheus metrics

adrianvielsack:
Adds a metric to export the last replication status per filesystem

### DIFF
--- a/daemon/job/active.go
+++ b/daemon/job/active.go
@@ -38,10 +38,11 @@ type ActiveSide struct {
 
 	prunerFactory *pruner.PrunerFactory
 
-	promRepStateSecs      *prometheus.HistogramVec // labels: state
-	promPruneSecs         *prometheus.HistogramVec // labels: prune_side
-	promBytesReplicated   *prometheus.CounterVec   // labels: filesystem
-	promReplicationErrors prometheus.Gauge
+	promRepStateSecs       *prometheus.HistogramVec // labels: state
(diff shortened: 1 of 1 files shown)


daemon/job/active.go
```diff
@@ -496,6 +506,19 @@ func (j *ActiveSide) do(ctx context.Context) {
 		replicationReport := j.tasks.replicationReport()
 		j.promReplicationErrors.Set(float64(replicationReport.GetFailedFilesystemsCountInLatestAttempt()))
 
+		for _, attempt := range replicationReport.Attempts {
```
problame, line 509: I think we only need the most recent attempt here, right?
Like so:

https://github.com/zrepl/zrepl/blob/83fdffbceff9d1c8b1690dc104f9cdbb7b2f8b87/replication/report/replication_report.go#L169-L176

daemon/job/active.go
```diff
@@ -496,6 +506,19 @@ func (j *ActiveSide) do(ctx context.Context) {
...
+			for _, filesystem := range attempt.Filesystems {
+				if filesystem.Info == nil {
```
problame, line 511: The "proper" way to check this is to inspect `attempt.State`, like so:

https://github.com/zrepl/zrepl/blob/83fdffbceff9d1c8b1690dc104f9cdbb7b2f8b87/replication/report/replication_report.go#L177-L191

problame (changes requested):
Sorry for the delay, I made a few comments!

PR #503, opened 2021-07-18
Allow resuming from unencrypted raw/compressed sends

infinisil:
I don't think there's any reason this check that prevented this (originally introduced in #259) should exist, so this PR removes it. I have tested this myself and can confirm that removing this check and doing an unencrypted raw send doesn't break anything.

### DIFF
--- a/replication/logic/replication_logic.go
+++ b/replication/logic/replication_logic.go
@@ -396,9 +396,7 @@ func (fs *Filesystem) doPlanning(ctx context.Context) ([]*Step, error) {
 		switch fs.policy.EncryptedSend {
 		case True:
 			encryptionMatches = resumeToken.RawOK && resumeToken.CompressOK
-		case False:
-			encryptionMatches = !resumeToken.RawOK && !resumeToken.CompressOK
-		case DontCare:
+		case False, DontCare:
 			encryptionMatches = true
 		}
 
--- a/zfs/zfs.go
+++ b/zfs/zfs.go
@@ -799,18 +799,9 @@ func (a ZFSSendArgsUnvalidated) validateEncryptionFlagsCorrespondToResumeToken(c
 		return gen.fmt("resume token `toguid` != expected: %v != %v", t.ToGUID, a.To.GUID)
 	}
 
-	if a.Encrypted.B {
-		if !(t.RawOK && t.CompressOK) {
-			return ZFSSendArgsResumeTokenMismatchEncryptionNotSet.fmt(
-				"resume token must have `rawok` and `compressok` = true but got %v %v", t.RawOK, t.CompressOK)
-		}
-		// fallthrough
-	} else {
-		if t.RawOK || t.CompressOK {
-			return ZFSSendArgsResumeTokenMismatchEncryptionSet.fmt(
-				"resume token must not have `rawok` or `compressok` set but got %v %v", t.RawOK, t.CompressOK)
-		}
-		// fallthrough
+	if a.Encrypted.B && !(t.RawOK && t.CompressOK) {
+		return ZFSSendArgsResumeTokenMismatchEncryptionNotSet.fmt(
+			"resume token must have `rawok` and `compressok` = true but got %v %v", t.RawOK, t.CompressOK)
 	}
 
 	return nil


Infinisil:
For searchability, this is the error you get without this patch after a send has been interrupted with `compress = true`:
```
tank/data PLANNING-ERROR (step 0/0, 0 B/0 B) server error: validate send arguments:
  resume token must not have `rawok` or `compressok` set but got false true
```

problame:
:thinking: interesting case. Could you show the `send` part of your `zrepl.yml`?

zfs/zfs.go
```diff
@@ -799,17 +799,9 @@ func (a ZFSSendArgsUnvalidated) validateEncryptionFlagsCorrespondToResumeToken(c
...
-		// fallthrough
-	} else {
-		if t.RawOK || t.CompressOK {
-			return ZFSSendArgsResumeTokenMismatchEncryptionSet.fmt(
-				"resume token must not have `rawok` or `compressok` set but got %v %v", t.RawOK, t.CompressOK)
-		}
+	if a.Encrypted.B && !(t.RawOK && t.CompressOK) {
+		return ZFSSendArgsResumeTokenMismatchEncryptionNotSet.fmt(
+			"resume token must have `rawok` and `compressok` = true but got %v %v", t.RawOK, t.CompressOK)
 		// fallthrough
```
problame, line 805: I think that `//fallthrough` ought to e removed.

problame (changes requested):
Apart from the stale comment, LGTM. Requesting review from @InsanePrawn

Infinisil:
> thinking interesting case. Could you show the `send` part of your `zrepl.yml`?

That would be
```yaml
send:
  compress: true
  embedded_data: true
```

Though as you could guess, it also doesn't work with `raw: true` (which I since switched to)

> Apart from the stale comment, LGTM. Requesting review from @InsanePrawn

The stale comment?

zfs/zfs.go
Infinisil: Done that :)

problame:
The remaining open question is: is it considered a security risk if an attacker can trigger an encrypted send even though the zrepl config says the dataset is unencrypted?

InsanePrawn:
LGTM.

The question goes further though - should there be a similar check for `send: { raw: true }`?

problame:
IIRC the problem is that some of the send flags don't show up in the resume token (?)
I think the following snippet from libzfs shows what's encoded in the token:

https://github.com/openzfs/zfs/blob/8f11b1d26efbdf5d654dfee6415518705b0ddf55/lib/libzfs/libzfs_sendrecv.c#L1648-L1673

In particular, I don't see a field for `properties`. Probably because properties are sent in the preliminary record for which recv doesn't create resume state.

InsanePrawn:
@Infinisil Since you switched to `raw`, would you expect a non-raw resume token to be used when `send: { raw: true }` is configured? Or should it discarded in favour of a fresh (possibly incremental) send?
This would be important to defend against sending plaintext from encrypted datasets, even if the user opted for `send: { raw: true, encrypted: false }`, for example to mix encrypted and unencrypted datasets in a job.

Infinisil:
I'd think a non-raw token should also be accepted in that case, because then the user still has the choice of discarding the resume token manually with `zfs receive -A <dataset>`. If non-raw tokens aren't accepted, zrepl would have to do this on its own, and give the user no choice to not do it.

InsanePrawn:
If we're speaking about defending against an attack, there would possibly be no resume token to discard since an attacker might be generating one on the fly.
To my mind, the value of defending against this outweighs the benefit of possibly resuming from a mismatched resume token, a situation that shouldn't normally arise anyway.

I'd even argue that discarding it is the more robust default behavior.

Should we make the behavior configurable, aka another case for #408?

Infinisil:
Ah yeah I've only thought about it from a users perspective. Ultimately I think it doesn't matter much and it should be fine to always discard mismatched resume tokens, especially if that could prevent some security issues. It's only a transient state after all, which should only happen when the `send` config is changed. So I also don't think making this configurable is necessary. If configurability is later desired by some users it can still be added.

problame:
Let's do this! @Infinisil do you want to take a stab at it?

Infinisil:
So the idea is to ensure that `raw`, `compressed`, `embedded_data` and `large_block` of the token are checked beforehand, and if they don't match zrepl's `send` config, the token is deleted with `zfs receive -A` and a full transfer is started?

And that independent of whether encryption is enabled. Encryption just has this additional check that `send.encrypted` is enabled in order to allow `raw` or `compressed`.

If I had more time to spend I'd go for it, but unfortunately I'm quite busy recently :disappointed:

problame:
Pretty much that, except

> ... deleted with zfs receive -A and a full transfer is started?

More precisely, the *step* is restarted from the beginning.

> And that independent of whether encryption is enabled. Encryption just has this additional check that send.encrypted is enabled in order to allow raw or compressed.

the other way around: `send.encrypted IMPLIES (fs needs to be encrypted + token needs to be raw+compressed)`

problame:
FWIW, the `valCtx.encEnabled` already contains the result of the check whether encryption is supported for the FS:

https://github.com/zrepl/zrepl/blob/5c6d69a69cbc4961a5779a5cca9200b25eb33b80/zfs/zfs.go#L624-L633

What's left to do is to bubble up a discernible error so that the call site in `endpoint.go` doesn't use the full send:

https://github.com/zrepl/zrepl/blob/0ceea1b792118453d792a8e25ecacdc8d8284ae1/endpoint/endpoint.go#L193-L196

zfs/zfs.go
```diff
@@ -799,18 +799,9 @@ func (a ZFSSendArgsUnvalidated) validateEncryptionFlagsCorrespondToResumeToken(c
...
-				"resume token must have `rawok` and `compressok` = true but got %v %v", t.RawOK, t.CompressOK)
-		}
-		// fallthrough
-	} else {
-		if t.RawOK || t.CompressOK {
-			return ZFSSendArgsResumeTokenMismatchEncryptionSet.fmt(
-				"resume token must not have `rawok` or `compressok` set but got %v %v", t.RawOK, t.CompressOK)
-		}
-		// fallthrough
+	if a.Encrypted.B && !(t.RawOK && t.CompressOK) {
```
InsanePrawn, line 802:
```suggestion
	if (a.Encrypted.B || a.Raw) && !(t.RawOK && t.CompressOK) {
```
should address the plaintext exfiltration issue imo.

@problame do we want/need to also check that -L and -e match? My gut feeling is no.

Do we need to rename/duplicate ZFSSendArgsResumeTokenMismatchEncryptionNotSet to ZFSSendArgsResumeTokenMismatchRawNotSet?

problame: Your proposed change returns the error unnecessarilty for the regular case:

```
token
  rawok = true
  compressor = false
config
  encrypted = false
  raw = true
  compress = false
```

problame: Re gut feeling: I don't think the other flags are relevant for security, but consider the following use case:
* new user sets up zrepl without any `send` options
* starts initial replication
* thinks "that'll take too long", interrupts initial replication
  * zfs saves resume token with (`compress = false, large_blocks = false`)
* user enables those features in the config, restarts daemon
* user expects that transfer will use `compress` and `large_blocks` now; they probably don't know and shouldn't care about the intricacies of the resume token

InsanePrawn: > Your proposed change returns the error unnecessarilty for the regular case:

why is this different from how `encrypted: true` behaves, i.e. why is compression requested on top of `-w` (raw) with `encrypted` but not `raw`? Those configs should produce the same send flags and tokens imo.

problame: > why is compression requested on top of -w (raw) with encrypted but not raw?

That's just how it is with zfs, i.e., if you specify `send --raw` then the resume token is `rawok=true; compressor=true`

problame: (that is, for encrypted datasets. not sure about unencrypted datasets)

Infinisil:
I'm not sure how this didn't happen before, but I seem to be running into a very similar error now with this patch:
```
resume token `rawok`=false and `compressok`=true are incompatible with encryption policy=false
```

This is odd because I have exactly only `raw = true` set for this job and not `compress`. This is on a `push` type.

I've had to manually `sudo zfs receive -A` on the sink to fix it.

problame:
Next time you hit that situation, do a `zfs send -nvt TOKEN` on the sender. This will dump the state encoded in the resume token. zrepl parses it in this file: https://github.com/zrepl/zrepl/blob/master/zfs/resume_token.go

Infinisil:
@problame Alright it happened again, here's the output from that

```
❯ zfs send -nvt 1-153e609dbf-138-789c636064000310a501c49c50360710a715e5e7a69766a63040c1e9c34eb3f6c7bffaac0064b323a9cb4fca4a4d2e616070dacc0e5687219f96569c5a02921181cbb321c9275596a41603e91d6b8e63d55f920f71859f94a6744e79c9020f24794eb07c5e626e2a904ecccb36d22fcacf2fd14f492c49d4cfc8cf4d75c8c94f4ecc893732303234b030348c37b430b5343488373030606090809a0ff37f6a6e526a4a7e36d87e6e24f1e4fcdc82a2d4e262a0141c00003fc831c7
resume token contents:
nvlist version: 0
	fromguid = 0xf3ea5fbf9a42c3cb
	object = 0x7b342
	offset = 0x140000
	bytes = 0xc7acb8
	toguid = 0xa074776c1b291a4e
	toname = tank2/root/data/home@local_20210811_185910_000
	embedok = 1
	compressok = 1
send from tank2/root/data/home@local_20210811_185410_000 to tank2/root/data/home@local_20210811_185910_000 estimated size is 3.56M
total estimated size is 3.56M
```

problame:
Ok, and what is the send-side zrepl config? Plus, what was the exact commit you were running (GH link please)?

Infinisil:
sink-side config:
```yaml
jobs:
- name: backups
  root_fs: tank/backup
  serve:
    clients:
      10.99.2.2: vario
    listen: :8888
    type: tcp
  type: sink
```

push-side config:
```yaml
jobs:
- name: data-snaps
  filesystems:
    tank2/root/data<: true
  pruning:
    keep:
    - regex: .*
      type: regex
  snapshotting:
    interval: 5m
    prefix: local_
    type: periodic
  type: snap
- name: data-push
  connect:
    address: 10.99.2.1:8888
    type: tcp
  filesystems:
    tank2/root/data<: true
  pruning:
    keep_receiver:
    - grid: 1x1d(keep=all) | 7x1d | 5x7d | 12x31d
      regex: .*
      type: grid
    keep_sender:
    - grid: 12x5m(keep=all) | 4x15m | 24x1h | 10x1d
      regex: .*
      type: grid
  send:
    raw: true
  snapshotting:
    interval: 1h
    prefix: repl_
    type: periodic
  type: push
```

Zrepl version: 0.4.0 with the commit from this PR cherry-picked
Commit of that: https://github.com/Infinisil/zrepl/commit/325cc5e7fb8c6d1d0ebe98b72a39007ab8b0b1fb

problame:
So `zfs send -nvt` reports that `rawok` is not set despite `send.raw = true`? :thinking:

replication/logic/replication_logic.go
```diff
@@ -396,9 +396,7 @@ func (fs *Filesystem) doPlanning(ctx context.Context) ([]*Step, error) {
 		switch fs.policy.EncryptedSend {
 		case True:
 			encryptionMatches = resumeToken.RawOK && resumeToken.CompressOK
-		case False:
-			encryptionMatches = !resumeToken.RawOK && !resumeToken.CompressOK
-		case DontCare:
+		case False, DontCare:
```
problame, line 399: @Infinisil any particular reason for this change here? I think we always set `policy.EncryptedSend` to `True` or `False` at this time, maybe we should instead expose `DontCare` in the config?

problame:
I would like to get this PR over the finish line.
It seems like the problem is that zrepl currently doesn't distinguish between a resume token that will do a *raw-but-unencrypted* send vs. an encrypted send. Correct?

helamonster:
Please correct me if I'm wrong, but is the planned change to force a resumable send to be aborted and re-sent from scratch!!!? 
I sure hope not. I sometimes have ~ 10 TiB steps that don't always finish, which effectively leaves the system permanently out of sync (until a manual send/recv is done) -- Or at least could end up wasting days or weeks of time re-sending data that's already on the receiver...

problame:
> Please correct me if I'm wrong, but is the planned change to force a resumable send to be aborted and re-sent from scratch!!!?

Definitely not if the config didn't change on sender/receiver. Did you spot that behavior in this PR?

problame:
> I would like to get this PR over the finish line.

That is still true. I'll spend some time reviewing this now, and see what it takes to finish.
@infinisil are you still willing to iterate on this PR, or should someone else take over?

problame:
Regarding the question of whether a given zfs send flag `$X` implies resume token `${X}ok`, I think this function is insightful:

* https://github.com/openzfs/zfs/blob/6c3c5fcfbe27d9193cd131753cc7e47ee2784621/module/zfs/dmu_send.c#L1947-L2012
* https://github.com/openzfs/zfs/blob/6c3c5fcfbe27d9193cd131753cc7e47ee2784621/module/zfs/dmu_recv.c#L877-L891
* https://github.com/openzfs/zfs/blob/6c3c5fc/lib/libzfs/libzfs_sendrecv.c#L1663-L1672

Basically, some zfs send flags make the DMU send code set some DMU send stream featureflags, although it's not a pure mapping, i..e, which DMU send stream flags are used depends somewhat on the dataset (e.g., is it encrypted or not, or, does it use zstd or not). 

Then, the receiver looks at some (but not all) feature flags and maps them to  `${X}ok` dataset zap attributes.

These are funnelled back to the sender 1:1 through the resume_token.

And the sender turns them into lzc flags.

So, the invariant that zfs send/recv should be implementing here is that the resume token will result in the same lzc flags.

What's zrepl's role in this?
We provide a virtual `encrypted` sendflag that is basically `raw`, but ensures that we only send encrypted datasets.

Anything else, we shouldn't care, because it's a futile effort to keep up with what ZFS does.

I'll simplify the zrepl code even further and be back with a patch within a week.
Sorry for taking so long to get to this insight :/

problame:
superseded by https://github.com/zrepl/zrepl/pull/613

Issue #497, opened 2021-06-26
`zrepl status` only reports half the true transfer rate


Malvineous:
I am copying a new filesystem for the first time, and I notice that the speed indicator seems quite wrong.  My other monitoring software is reporting between 300-600 MB/sec however `zrepl status` is only showing 90-150 MiB/s:

    Started: 2021-06-26 19:53:37 +1000 AEST (lasting 22m51s)
    Progress: [=|-------------------------------------------------] 382.5 GiB / 10.1 TiB @ 149.3 MiB/s

Unless my maths are wrong, 22m51s is 1371 seconds, and 382.5 GiB in 1371 seconds is 285 MiB/s.  zrepl seems to be reporting only half the actual speed.  Sure enough if I look at the actual filesystem values reported by zrepl, they are increasing by around 0.3 GiB/sec, double the reported speed.

I tried to find in the code where this is calculated and it looks like it might be at https://github.com/zrepl/zrepl/blob/master/client/status/viewmodel/bytesprogresshistory.go#L42 but there's some weird factor in play, so I'm wondering whether this is making the calculation come out wrong?  It looks like it's designed to gradually creep up to the true speed but despite leaving it for 10 minutes, it only ever shows around half the true speed.

Malvineous:
I tried experimenting with the `factor` value and `1.0` gives me an accurate speed reading but anything less than this and the displayed transfer speed is consistently lower than what it really is.

I guess because it starts off at `0.0` and that value is carried forward forever thanks to the factor, it never recovers and never reaches the true speed.

problame:
It's an exponentially moving average which we update every time we poll.
The poll interval is controlled via this variable (default: 1s)

https://github.com/zrepl/zrepl/blob/master/client/status/status.go#L29

Here's the expected behavior for different factors.

![image](https://user-images.githubusercontent.com/956573/123623812-2c576980-d80e-11eb-87f2-d75dfecebb30.png)

If you poll faster (i.e., change `--delay`), the faster updates will compress the curve on the x-axis.

Action items:

* experiment with picking a higher factor (you already did that)
* choose the factor depending on `--delay` to compensate for the compression/dilation of the x-axis
  * More precisely, don't use `--delay` but the actual `deltaT` to compute the weighing

There's also #339  (rate-limiting), so we'll eventually need to be more precise about transfer rates and move the logic into the daemon. But in the meantime, we can tackle the items above easily.

[...]

PR #506, opened 2021-08-06
job-level bandwidth limit

### LINKED ISSUE #339: Transport rate limiting
I have some 100TB raidz filesystems which I currently replicate with syncoid, using ssh-hpn with encryption disabled so as not to be a bottleneck.

If I let them run at full speed they achieve ~550MiB/sec, but put so much load on the source server that they impact client performance (and sometimes crash the HBA!)  But if I limit them to 200MiB/sec then everything is fine.

With syncoid I specify `--source-bwlimit=200m`

zrepl would be a much cleaner replacement, as its native TCP transport removes the need for ssh-hpn, but I still need some way to limit the bandwidth.

The workarounds I can think of are:

- at the sender side, point at a TCP proxy on localhost which implements rate limiting
- using iptables rate limiting

problame:
This PR adds the ability to configure a job-level bandwidth / data rate limit, based on a token bucket.
The limit is applied _above_ the `transport`, and only to the `zfs send` / `zfs recv` operation.
This fixes #339.

Also, the data rate display in `zrepl status` can average towards the wrong value (#497). 
This PR includes a workaround that disables the averaging.

Todo

- [x] Testing in VMs
- [x] Testing in production deployment (considered very low risk)
  - Determine trade-offs with  around with different values for `bucket_capacity`
  - Watch out for CPU overhead
    - Check 1: if no bandwidth limit is configured, the overhead should be practically zero.
    - Check 2: Determine CPU overhead in Multi-GiB/s workloads
      - Hardware information (`cat /proc/cpuinfo` or similar)
      - zrepl job's `transport.type`
      - Average CPU utilization in `%` before and after enabling a bandwidth limit
      - If the CPU utilization is problematic, the data rate as reported by `zrepl status` / `iftop` or similar tools
- [x] Better commit message
- [x] Sponsor attribution at https://zrepl.github.io/supporters.html
- [x] ~~Replace `zrepl status` workaround with actual fix for data rate averaging~~ Rename `zrepl status` workaround commit.

### DIFF
--- a/docs/configuration/sendrecvoptions.rst
+++ b/docs/configuration/sendrecvoptions.rst
@@ -36,6 +36,9 @@ See the `upstream man page <https://openzfs.github.io/openzfs-docs/man/8/zfs-sen
     * - ``encrypted``
       -
       - Specific to zrepl, :ref:`see below <job-send-options-encrypted>`.
+    * - ``bandwidth_limit``
+      -
+      - Specific to zrepl, :ref:`see below <job-send-recv-options-bandwidth-limit>`.
     * - ``raw``
       - ``-w``
       - Use ``encrypted`` to only allow encrypted sends.
@@ -138,6 +141,7 @@ Recv Options
          override: {
            "org.openzfs.systemd:ignore": "on"
          }
+       bandwidth_limit: ... # see below
      ...
 
 .. _job-recv-options--inherit-and-override:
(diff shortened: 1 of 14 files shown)


docs/configuration/sendrecvoptions.rst
```diff
@@ -212,3 +216,26 @@ and property replication is enabled, the receiver must :ref:`inherit the followi
...
+Common Options
+~~~~~~~~~~~~~~
+
+.. _job-send-recv-options-bandwidth-limit:
+
+Bandwidth Limit (send & recv)
+-----------------------------
+
+Both ``send`` and ``recv`` can be limited to a maximum bandwidth through ``bandwidth_limit``.
+For most users, it should be sufficient to just set ``bandwidth_limit.max``.
```
InsanePrawn, line 229: still we need some hint or optional explanation on what a bucket is and what changing its size does.

PR #512, opened 2021-09-05
Fixes #410 - Uses zstream token.

### LINKED ISSUE #410: use `zstream token` to decode resume token
Main benefit would be that we'd be using a supported tool instead of dry-sends.
see https://github.com/openzfs/zfs/commit/02fced3067798cbc5472f180e4c6d36f4c441a77

### LINKED ISSUE #510: zfs send resume issues - Running on wrong context / Failing to decode valid tokens
I'm going to be up front and state I haven't done a lot of detailed testing for this: I just figured it out from turning on audit logging on both servers for any calls to ZFS and realized what was happening.

I have a backup server that pulls from sources to itself. It usually works fine, but I've noticed sometimes if a connection drops or something happens, the receives always fail with "zfs binary does not allow decoding resume token or zrepl cannot scrape zfs output". Normally, I just used zfs recv -A, but today I decided to REALLY dig into things. Turning on debug logging didn't help: I got the calls, and confirmed they were correct, but they worked just fine on the source. They also worked when using zrepl test decoderesumetoken, so I couldn't figure out what was going on.

As said in the first line however, I suddenly got the idea to just turn on auditing for zfs (auditctl -a exit,always -S execve -F path=/usr/sbin/zfs), and to my surprise I didn't see any hits for nvt in audit.log: It wasn't calling it on the right server. Trying the same command to turn on audit logging on the destination server revealed it was indeed trying to call zfs send -nvt (the token) on the local server: Which of course fails because the snapshot it's trying to start the send from doesn't exist on the local server (Or even the pool, usually).

Issue #410 MAY resolve this, but then looking at the code it should actually still work unless there's some kind of an error in the [resume token error checking code](https://github.com/zrepl/zrepl/blob/3b5a1a8b9a14c167d1af3ab7af115d5f9ffe9c2f/zfs/resume_token.go#L188), but go is just one of those languages I haven't gotten around to learning yet: It does however return an error (See below)
```
clio:~# zfs send -nvt 1-139fd78ae3-110-789c636064000310a501c49c50360710a715e5e7a69766a6304081dfd3e667ede76f052a00d9ec48eaf293b252934b18189ebc83a8c3904f4b2b4e2d01c908c1e5d990e4932a4b528b817486bf3d2336fd25f9105
75c739af3adcc728d7f04923c27583e2f3137154867a4e66616eba6e42767a716e94328dd94c49244fd82fce292f4a2d46287aaa2d4829c7823032343030b6383780343234323a378030303981f01eb673158
resume token contents:
nvlist version: 0
        fromguid = 0x51dacf87e683e54e
        object = 0xeee4
        offset = 0x120000
        bytes = 0x13f4f68
        toguid = 0x4fac3976f69c42d6
        toname = themis-docker/docker-data/postgres@zrepl_20210830_012122_000
cannot resume send: 'themis-docker/docker-data/postgres@zrepl_20210830_012122_000' used in the initial send no longer exists
clio:~# zrepl test decoderesumetoken --token 1-139fd78ae3-110-789c636064000310a501c49c50360710a715e5e7a69766a6304081dfd3e667ede76f052a00d9ec48eaf293b252934b18189ebc83a8c3904f4b2b4e2d01c908c1e5d990e4932a4b528b
817486bf3d2336fd25f910575c739af3adcc728d7f04923c27583e2f3137154867a4e66616eba6e42767a716e94328dd94c49244fd82fce292f4a2d46287aaa2d4829c7823032343030b6383780343234323a378030303981f01eb673158
zfs binary does not allow decoding resume token or zrepl cannot scrape zfs output
clio:~#
```

I assume from the current code the easiest solution would be to actually check if a token is valid by sending the check request to the destination context, or as said earlier using zstream token command in #410. I might give that a shot later since from what I can see it wouldn't be too hard: A few changes to the regex and the actual command should be all that's required.

ShaRose:
This is a quick fix, but it does work on my system to resume stuck transfers (And thus also fixes #510 albeit not using the best solution). You can actually run the example tokens through zrepl test decoderesumetoken without errors, which is a nice plus. That could possibly be added to platform testing or even build testing later, and we can even add both checks: This (or manual decoding) as a first pass sanity check, then running on the other side for flag information and further validation that the tokens are good. But for now this is, as said, a quick fix.

Changes:

Removed resumeTokenContentsRE and accompanying code (As zstream token doesn't have a starting string to regex against).

Updated resumeTokenNVListRE to match the new format for zstream token.

Updated some comments.

Use zstream token instead of zfs send -nvt.

Commented out some of the other optional fields this returns as they are never in the actual token: There isn't much of a reason to keep the dead code there, but thought it should be clear that those fields are not used as is.

### DIFF
--- a/zfs/resume_token.go
+++ b/zfs/resume_token.go
@@ -28,9 +28,8 @@ type ResumeToken struct {
 	HasSavedOk, SavedOk           bool
 }
 
-var resumeTokenNVListRE = regexp.MustCompile(`\t(\S+) = (.*)`)
+var resumeTokenNVListRE = regexp.MustCompile(` *(\S+): (.*)`)
 
-var resumeTokenContentsRE = regexp.MustCompile(`resume token contents:\nnvlist version: 0`)
 var resumeTokenIsCorruptRE = regexp.MustCompile(`resume token is corrupt`)
 
 var ResumeTokenCorruptError = errors.New("resume token is corrupt")
@@ -154,7 +153,7 @@ func ResumeRecvSupported(ctx context.Context, fs *DatasetPath) (bool, error) {
 	return poolSup.supported, nil
 }
 
-// Abuse 'zfs send' to decode the resume token
+// Abuse 'zstream token' to decode the resume token
 //
 // FIXME: implement nvlist unpacking in Go and read through libzfs_sendrecv.c
 func ParseResumeToken(ctx context.Context, token string) (*ResumeToken, error) {
@@ -173,38 +172,30 @@ func ParseResumeToken(ctx context.Context, token string) (*ResumeToken, error) {
 	// From an incremental send
 	// 1-c49b979a2-e0-789c636064000310a501c49c50360710a715e5e7a69766a63040c1eabb735735ce8f8d5400b2d991d4e52765a5269740f82080219f96569c5ac2000720793624f9a4ca92d46206547964fd25f91057f09e37babb88c9bf5503499e132c9f97989bcac050909f9f63a80f34abc421096616007c881d4c
 
-	// Resulting output of zfs send -nvt <token>
+	// Resulting output of zstream token <token>
 	//
-	//resume token contents:
-	//nvlist version: 0
-	//	fromguid = 0x595d9f81aa9dddab
-	//	object = 0x1
-	//	offset = 0x0
-	//	bytes = 0x0
-	//	toguid = 0x854f02a2dd32cf0d
-	//	toname = pool1/test@b
-	//cannot resume send: 'pool1/test@b' used in the initial send no longer exists
+	//     fromguid: 6439478421471747499
+	//     object: 1
+	//     offset: 0
+	//     bytes: 0
+	//     toguid: 9605899428723609357
+	//     toname: 'pool1/test@b'
 
-	cmd := zfscmd.CommandContext(ctx, ZFS_BINARY, "send", "-nvt", string(token))
+	cmd := zfscmd.CommandContext(ctx, "zstream", "token", string(token))
 	output, err := cmd.CombinedOutput()
 	if err != nil {
 		if exitErr, ok := err.(*exec.ExitError); ok {
 			if !exitErr.Exited() {
 				return nil, err
 			}
-			// we abuse zfs send for decoding, the exit error may be due to
-			// a) the token being from a third machine
-			// b) it no longer exists on the machine where
 		} else {
 			return nil, err
 		}
 	}
 
-	if !resumeTokenContentsRE.Match(output) {
-		if resumeTokenIsCorruptRE.Match(output) {
-			return nil, ResumeTokenCorruptError
-		}
-		return nil, ResumeTokenDecodingNotSupported
+	// zstream token doesn't really HAVE a normal starting string to regex against, thus why I'm only checking for a corrupt token.
+	if resumeTokenIsCorruptRE.Match(output) {
+		return nil, ResumeTokenCorruptError
 	}
 
 	matches := resumeTokenNVListRE.FindAllStringSubmatch(string(output), -1)
@@ -231,6 +222,7 @@ func ParseResumeToken(ctx context.Context, token string) (*ResumeToken, error) {
 			rt.HasToGUID = true
 		case "toname":
 			rt.ToName = val
+		/*
 		case "rawok":
 			rt.HasRawOk = true
 			rt.RawOK, err = strconv.ParseBool(val)
@@ -261,6 +253,7 @@ func ParseResumeToken(ctx context.Context, token string) (*ResumeToken, error) {
 			if err != nil {
 				return nil, ResumeTokenParsingError
 			}
+		*/
 		}
 	}
 


problame:
We need to maintain backwards-compatibility for the time being, so the change needs to try `zstream token` first, and if the binary does not exist, fall back to the `send -nvt` way.

problame:
We can't take this PR as is because

1. It still needs to be backwards-compatible with platforms that don't have `zstream`.
2. It fails the zrepl platformtests for resume token decoding

```
make test-platform-bin
sudo ./artifacts/platformtest-linux-amd64 \
	-poolname "zreplplatformtest" \
        -imagepath "/tmp/zreplplatformtest.pool.img" \
        -mountpoint "/tmp/zreplplatformtest.pool" \
	-run ResumeToken
```

Output:

```
...
2022-07-08T18:41:29Z [INFO][platformtest][Uv38$Uv38]: filesystem in resume token field `toname` = "'zreplplatformtest/rootds/send er" does not match expected value "zreplplatformtest/rootds/send er"
2022-07-08T18:41:29Z [ERRO][platformtest][Uv38$Uv38]:
        Error Trace:    sendArgsValidation.go:132
                                                harness.go:194
                                                harness.go:195
                                                harness.go:108
                                                harness.go:43
                                                proc.go:255
                                                asm_amd64.s:1581
        Error:          Not equal:
                        expected: 4
                        actual  : 2

panic while running test: platformtest: FailNow called on context
main.runTestCase.func1.1
        /home/cs/zrepl/zrepl/platformtest/harness/harness.go:191
runtime.gopanic
        /home/cs/go1.17.5/src/runtime/panic.go:1038
github.com/zrepl/zrepl/platformtest.(*Context).FailNow
        /home/cs/zrepl/zrepl/platformtest/platformtest.go:34
github.com/zrepl/zrepl/platformtest.(*Context).Errorf
        /home/cs/zrepl/zrepl/platformtest/platformtest.go:30
github.com/stretchr/testify/assert.Fail
        /home/cs/go/pkg/mod/github.com/stretchr/testify@v1.6.1/assert/assertions.go:262
github.com/stretchr/testify/assert.Equal
        /home/cs/go/pkg/mod/github.com/stretchr/testify@v1.6.1/assert/assertions.go:346
github.com/stretchr/testify/require.Equal
        /home/cs/go/pkg/mod/github.com/stretchr/testify@v1.6.1/require/require.go:162
github.com/zrepl/zrepl/platformtest/tests.SendArgsValidationResumeTokenEncryptionMismatchForbidden
        /home/cs/zrepl/zrepl/platformtest/tests/sendArgsValidation.go:132
main.runTestCase.func1
        /home/cs/zrepl/zrepl/platformtest/harness/harness.go:194
main.runTestCase
        /home/cs/zrepl/zrepl/platformtest/harness/harness.go:195
main.HarnessRun
        /home/cs/zrepl/zrepl/platformtest/harness/harness.go:108
main.main
        /home/cs/zrepl/zrepl/platformtest/harness/harness.go:43
runtime.main
        /home/cs/go1.17.5/src/runtime/proc.go:255
runtime.goexit
        /home/cs/go1.17.5/src/runtime/asm_amd64.s:1581
2022-07-08T18:41:29Z [DEBG]: begin executing command="\"zpool\" [\"export\" \"zreplplatformtest\"]"
2022-07-08T18:41:29Z [DEBG]: command output command="\"zpool\" [\"export\" \"zreplplatformtest\"]" output=""
2022-07-08T18:41:29Z [DEBG]: done executing command="\"zpool\" [\"export\" \"zreplplatformtest\"]"
TEST FAILED

PASSING TESTS:
  github.com/zrepl/zrepl/platformtest/tests.SendArgsValidationResumeTokenDifferentFilesystemForbidden
SKIPPED TESTS: []
FAILED TESTS:
  github.com/zrepl/zrepl/platformtest/tests.ResumeTokenParsing
  github.com/zrepl/zrepl/platformtest/tests.SendArgsValidationResumeTokenEncryptionMismatchForbidden
```

Looking at the difference in output, we see that `zstream` adds single-quotes around `toname`:

```
root@zrepl-1-2204:[/]: zstream token 1-c9cc2b0aa-d8-789c636064000310a500c4ec50360710e72765a52697303030419460caa7a515a79680647ce0f26c48f2499525a9c540bac3d20fabfe92fcf4d2cc140606179d4587e62db9c5eb8124cf0996cf4bcc4d6560a82a4a2dc829c8492c49cb2fca051a59a25f949f5f9252ac5f9c9a97a2905ae490a8509c9758509c915fc20003005abd23c7
     object: 2
     offset: 4980736
     bytes: 5126536
     toguid: 998291269193182276
     toname: 'zreplplatformtest/rootds/send er@a snapshot'
```

```
root@zrepl-1-2204:[/]: zfs send -nvt 1-c9cc2b0aa-d8-789c636064000310a500c4ec50360710e72765a52697303030419460caa7a515a79680647ce0f26c48f2499525a9c540bac3d20fabfe92fcf4d2cc140606179d4587e62db9c5eb8124cf0996cf4bcc4d6560a82a4a2dc829c8492c49cb2fca051a59a25f949f5f9252ac5f9c9a97a2905ae490a8509c9758509c915fc20003005abd23c7
resume token contents:
nvlist version: 0
        object = 0x2
        offset = 0x4c0000
        bytes = 0x4e3988
        toguid = 0xddaa49ec2a22c44
        toname = zreplplatformtest/rootds/send er@a snapshot
cannot resume send: 'zreplplatformtest/rootds/send er@a snapshot' used in the initial send no longer exists
```

The reason is that `zstream` uses `dump_nvlist` whereas `zfs send -nvt` hand-prints the output.

https://github.com/openzfs/zfs/blob/cb01da68057dcb9e612e8d2e97d058c46c3574af/cmd/zstream/zstream_token.c#L73

https://github.com/openzfs/zfs/blob/cb01da68057dcb9e612e8d2e97d058c46c3574af/lib/libzfs/libzfs_sendrecv.c#L1708-L1715

---

In my opinion, `zstream` should just use the existing functionality to dump JSON.
Then, switching to `zstream` would provide a substantial simplification.
Right now, I find it less appealing.

Considering the age of this PR and lack of responses, I'll close it, and close the underlying issue #410.

If someone wants to implement JSON dumping for zstream in OpenZFS, we can have a new issue and a new PR.

Issue #516, opened 2021-09-13
Improve error detection in Prometheus metrics


lapo-luchini:
Hi, I have a zrepl node which is exporting this:

`zrepl_replication_filesystem_errors{zrepl_job="backup"} 0`

but is actually been failing synchronizations for one month, the underlying problem might be related to #495 but I was thinking, in the meantime, it could be useful (and easy) to add a metric such as:

`zrepl_replication_filesystem_last_complete{zrepl_job="backup"} 1631543787665`

So that I can use a more resilient alerting such as "last complete was sent more than 24h ago".

problame:
Yes, that sounds reasonable. Could you describe in what way the replications failed?

Also, cc #481

lapo-luchini:
If a system to detect a stalled connection (i.e. `0.0 KiB` for 5 minutes, for example) was in place, the new metric I propose wouldn't be needed anymore and it would actually solve the problem |(and retry automatically later), not only give a metric to warn me to go and reset zrepl manually. 🤔
…but that's a whole different RFE (and, I guess, a lot more complex to implement).

lapo-luchini:
Nice, thanks!
I'd like to continue that patch to add "per-filesystem" granularity, and potentially using the date of the last sent snapshot date (for that filesystem) instead of the date it was sent. I'll probably need some suggestions/help with that, though, as it gets more invasive of the existing code.

PR #530, opened 2021-10-20
Add "last completed" metric to ease error reporting.

lapo-luchini:
My own attempt at closing #516.

(Please notice I know no Go, I only tried to emulate the surrounding code.)

### DIFF
--- a/daemon/job/active.go
+++ b/daemon/job/active.go
@@ -42,6 +42,7 @@ type ActiveSide struct {
 	promPruneSecs         *prometheus.HistogramVec // labels: prune_side
 	promBytesReplicated   *prometheus.CounterVec   // labels: filesystem
 	promReplicationErrors prometheus.Gauge
+	promReplicationLastComplete prometheus.Gauge
 
 	tasksMtx sync.Mutex
 	tasks    activeSideTasks
@@ -321,12 +322,18 @@ func activeSide(g *config.Global, in *config.ActiveJob, configJob interface{}) (
 		Help:        "number of bytes replicated from sender to receiver per filesystem",
 		ConstLabels: prometheus.Labels{"zrepl_job": j.name.String()},
 	}, []string{"filesystem"})
-
 	j.promReplicationErrors = prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace:   "zrepl",
 		Subsystem:   "replication",
 		Name:        "filesystem_errors",
 		Help:        "number of filesystems that failed replication in the latest replication attempt, or -1 if the job failed before enumerating the filesystems",
+		ConstLabels: prometheus.Labels{"zrepl_job": j.name.String()},
+	})
+	j.promReplicationLastComplete = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   "zrepl",
+		Subsystem:   "replication",
+		Name:        "filesystem_last_complete",
+		Help:        "timestamp of last correctly completed replication",
 		ConstLabels: prometheus.Labels{"zrepl_job": j.name.String()},
 	})
 
@@ -360,6 +367,7 @@ func (j *ActiveSide) RegisterMetrics(registerer prometheus.Registerer) {
 	registerer.MustRegister(j.promPruneSecs)
 	registerer.MustRegister(j.promBytesReplicated)
 	registerer.MustRegister(j.promReplicationErrors)
+	registerer.MustRegister(j.promReplicationLastComplete)
 }
 
 func (j *ActiveSide) Name() string { return j.name.String() }
@@ -494,7 +502,11 @@ func (j *ActiveSide) do(ctx context.Context) {
 		repCancel()   // always cancel to free up context resources
 
 		replicationReport := j.tasks.replicationReport()
-		j.promReplicationErrors.Set(float64(replicationReport.GetFailedFilesystemsCountInLatestAttempt()))
+		var numErrors = replicationReport.GetFailedFilesystemsCountInLatestAttempt()
+		j.promReplicationErrors.Set(float64(numErrors))
+		if numErrors == 0 {
+			j.promReplicationLastComplete.SetToCurrentTime()
+		}
 
 		endSpan()
 	}
--- a/daemon/job/active.go
+++ b/daemon/job/active.go
@@ -42,7 +42,7 @@ type ActiveSide struct {
 	promPruneSecs         *prometheus.HistogramVec // labels: prune_side
 	promBytesReplicated   *prometheus.CounterVec   // labels: filesystem
 	promReplicationErrors prometheus.Gauge
-	promReplicationLastSuccessful prometheus.Gauge
+	promLastSuccessful    prometheus.Gauge
 
 	tasksMtx sync.Mutex
 	tasks    activeSideTasks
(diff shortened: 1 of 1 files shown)


lapo-luchini:
This could maybe improved by using "timestamp of the most recent correctly transferred snapshot" to have partial results during catch-up of a long sync with many snapshots to be sent… but if multiple filesystems are involved one could be behind the others so it would have to be done very carefully.
Current patch is very simplistic as it has missing value at start and only sets it on a successful sync, which is fine if all your servers are updated often, less nice if some is way behind and has been recently reset.
@problame what do you think about it?

lapo-luchini:
PS: is there a way to anonymously check CircleCI results? I don't have an account as I disagree to giving them _write_ access to all my _private_ repositories only to have read-only access on this PR report…

daemon/job/active.go
```diff
@@ -321,14 +322,20 @@ func activeSide(g *config.Global, in *config.ActiveJob, configJob interface{}) (
...
 		Namespace:   "zrepl",
 		Subsystem:   "replication",
 		Name:        "filesystem_errors",
 		Help:        "number of filesystems that failed replication in the latest replication attempt, or -1 if the job failed before enumerating the filesystems",
 		ConstLabels: prometheus.Labels{"zrepl_job": j.name.String()},
 	})
+	j.promReplicationLastComplete = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   "zrepl",
+		Subsystem:   "replication",
+		Name:        "filesystem_last_complete",
```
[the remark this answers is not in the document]
lapo-luchini: Done.

daemon/job/active.go
```diff
@@ -494,7 +502,11 @@ func (j *ActiveSide) do(ctx context.Context) {
 		repCancel()   // always cancel to free up context resources
 
 		replicationReport := j.tasks.replicationReport()
-		j.promReplicationErrors.Set(float64(replicationReport.GetFailedFilesystemsCountInLatestAttempt()))
+		var numErrors = replicationReport.GetFailedFilesystemsCountInLatestAttempt()
+		j.promReplicationErrors.Set(float64(numErrors))
+		if numErrors == 0 {
```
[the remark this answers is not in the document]
lapo-luchini: I take from your comment this is fine as it is, for now, right?

daemon/job/active.go
```diff
@@ -321,14 +322,20 @@ func activeSide(g *config.Global, in *config.ActiveJob, configJob interface{}) (
...
 		Namespace:   "zrepl",
 		Subsystem:   "replication",
 		Name:        "filesystem_errors",
 		Help:        "number of filesystems that failed replication in the latest replication attempt, or -1 if the job failed before enumerating the filesystems",
 		ConstLabels: prometheus.Labels{"zrepl_job": j.name.String()},
 	})
+	j.promReplicationLastSuccessful = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   "zrepl",
+		Subsystem:   "replication",
+		Name:        "filesystem_last_successful",
```
problame, line 335: One last nit. I think we should just use `Name: "last_successful"`. Or, we should make this a `GaugeVec` with a `filesystem` label. cc @adrianvielsack  / https://github.com/zrepl/zrepl/pull/481 , would a `GaugeVec` in this PR solve your use case as well?

lapo-luchini: Yeah, I don't need fs-level detail, but it would be a nice plus.
I'll take a stab at the patch.

lapo-luchini: Mhh, AFAICT that needs to go into `replication_logic.go` line 643 (just after `log.Debug("receive finished")`) but that involves changing a couple of structs and is a tad bit more complex than my grasp of Go allows.

lapo-luchini: What about both?
With the suggested rename (filesystem_last_successful→last_successful) this PR could land now, then I'll try and do the "each filesystem" patch, but it's way more complex and will take a bit.

lapo-luchini: Metric renamed as suggested. (I also renamed the variable because else a lot of lines would need new spaces, this was missing in previous commit)

lapo-luchini:
I'm using the following query to show this:
```PromQL
(time() - zrepl_replication_last_successful{job="$prom_job_name"}) * on(instance) group_left(raw) zrepl_version_daemon{job="$prom_job_name"}
```

Grafana panel:
```JSON
{"fieldConfig":{"defaults":{"custom":{"align":null,"filterable":false},"thresholds":{"mode":"absolute","steps":[{"color":"green","value":null},{"color":"#EAB839","value":7200},{"color":"red","value":86400}]},"mappings":[{"from":"315576000","id":1,"text":"never","to":"9999999999","type":2,"value":""}],"color":{"mode":"thresholds"},"unit":"s"},"overrides":[{"matcher":{"id":"byName","options":"last completed"},"properties":[{"id":"custom.displayMode","value":"color-text"}]}]},"gridPos":{"h":8,"w":24,"x":0,"y":13},"id":55,"options":{"showHeader":true,"sortBy":[{"desc":true,"displayName":"last completed"}]},"pluginVersion":"7.5.11","targets":[{"expr":"(time() - zrepl_replication_last_successful{job=\"$prom_job_name\"}) * on(instance) group_left(raw) zrepl_version_daemon{job=\"$prom_job_name\"}","legendFormat":"","interval":"","exemplar":true,"instant":true,"refId":"A"}],"title":"Last completed replication age","transformations":[{"id":"labelsToFields","options":{}},{"id":"organize","options":{"excludeByName":{"Time":true,"audience":true,"datacenter":false,"job":true},"indexByName":{"Time":0,"Value":7,"audience":1,"datacenter":2,"instance":3,"job":4,"raw":6,"zrepl_job":5},"renameByName":{"Value":"last completed","job":"","raw":"version","zrepl_job":""}}}],"type":"table","datasource":null}
```

problame:
I played around with this panel. For the record, it looks like this:

image


I would like to include it in the dist/grafana panel that we ship, but the problem is that the thresholds are hard-coded in the panel json. I guess we'd like to have 1x or 2x `$interval` for push side jobs, where `$interval` is the zrepl job's `interval` yaml value.

lapo-luchini:
BTW: I changed it a bit more, e.g. by adding a `max_over_time()` in order to avoid missing values after zrepl restarts before first sync is finished. Feel free to include it at your will, in original or modified form.
```
{"fieldConfig":{"defaults":{"custom":{"align":null,"filterable":false},"thresholds":{"mode":"absolute","steps":[{"color":"green","value":null},{"color":"#EAB839","value":7200},{"color":"red","value":86400}]},"mappings":[{"from":"315576000","id":1,"text":"never","to":"9999999999","type":2,"value":""}],"color":{"mode":"thresholds"},"unit":"s"},"overrides":[{"matcher":{"id":"byName","options":"last completed"},"properties":[{"id":"custom.displayMode","value":"color-text"}]}]},"gridPos":{"h":8,"w":24,"x":0,"y":13},"id":55,"options":{"showHeader":true,"sortBy":[{"desc":true,"displayName":"last completed"}]},"pluginVersion":"7.5.11","targets":[{"exemplar":true,"expr":"(time() - max_over_time(zrepl_replication_last_successful{job=\"$prom_job_name\"}[30d])) * on(instance) group_left(raw) zrepl_version_daemon{job=\"$prom_job_name\"}","instant":true,"interval":"","legendFormat":"","refId":"A"}],"title":"Last completed replication age","transformations":[{"id":"labelsToFields","options":{}},{"id":"organize","options":{"excludeByName":{"Time":true,"audience":true,"datacenter":false,"job":true},"indexByName":{"Time":0,"Value":7,"audience":1,"datacenter":2,"instance":3,"job":4,"raw":6,"zrepl_job":5},"renameByName":{"Value":"last completed","job":"","raw":"version","zrepl_job":""}}}],"type":"table","datasource":null      }
```

PR #535, opened 2021-11-12
docs: prune: Fix fall of snapshots into buckets for the example to really reflect right-exclusiveness

samy-mahmoudi:
I also applied other changes: the three oldest commits could be squashed / fixed up after review.


InsanePrawn:
LGTM, thanks!

problame:
Thanks, will squash & push to master.

Issue #551, opened 2021-12-23
[Question] Fan-out replication with 1 snapshot job and N source jobs?


chenxiaolong:
I'm looking to implement fan-out replication as described on: https://zrepl.github.io/configuration/overview.html#overview-how-replication-works. Specifically, I have 3 hosts, (`A`, `B1`, `B2`), and I want to replicate from A to both B1 and B2.

Assuming, I have a setup like this:
* 1 snapshot job on `A` that does the snapshotting and pruning
* 1 source job on `A` + 1 pull job on `B1` that does not prune the sender
* 1 source job on `A` + 1 pull job on `B2` that does not prune the sender
* Datasets being sent to `B1` and `B2` are the same (not disjoint)

Would there any complications with this setup?

My understanding from the documentation is that datasets being replicated don't have to be disjoint as long as only one job does the pruning. I also understand that N pull jobs with 1 source job is not supported because the client identity name is not part of the zfs abstraction naming scheme. It _seems_ like N pull jobs, each with a corresponding unique source job, would work fine given that that the source job name is encoded in `zrepl zfs-abstraction list`. Is my understanding correct?

Thank you!

problame:
(Disclaimer: I haven't tested this anywhere)
(Disclaimer 2: there are going to be some error-level log messages if the snapjob's pruner and the replication jobs race, but they should be safe to ignore)

> My understanding from the documentation is that datasets being replicated don't have to be disjoint as long as only one job does the pruning.

Yes.

> I also understand that N pull jobs with 1 source job is not supported because the client identity name is not part of the zfs abstraction naming scheme.

Correct. To expand this a little for future readers:
Pull jobs receive the source-side filesystems directly into `$root_fs`, so there's no client identity involved on the pulling = receiving side.
But: the replication cursor bookmark does not include the client identity of the pull job. So, multiple pull jobs would share the same replication cursor bookmark. Which will break incremental replication at some point.
That's a flaw in the design, and could be fixed.

> It seems like N pull jobs, each with a corresponding unique source job, would work fine given that that the source job name is encoded in zrepl zfs-abstraction list. Is my understanding correct?

Yes, spot-on.

---

Please report your experience, and maybe create a [quickstart guide variant](https://zrepl.github.io/quickstart.html#configuration-examples) for your use case. If you don't know RST, write in in markdown and make a PR, I'll format it.

chenxiaolong:
[...]

PR #552, opened 2021-12-31
docs: Add quick-start guide for fan-out replication

chenxiaolong:
This PR adds adds a quick-start guide for fan-out replication using `source`/`pull` jobs. The `More Than 2 Machines` section in the configuration overview page has also been updated to indicate that this type of setup is possible.

See: #551

### DIFF
--- a/docs/configuration/overview.rst
+++ b/docs/configuration/overview.rst
@@ -293,6 +293,24 @@ This section might be relevant to users who wish to *fan-in* (N machines replica
   * ``sink`` constrains each client to a disjoint sub-tree of the sink-side dataset hierarchy ``${root_fs}/${client_identity}``.
     Therefore, the different clients cannot interfere.
 
+.. _fan-out-replication:
+
+* Fan-out replication: 1 ``pull`` job on each target server, each with a corresponding unique ``source`` job on the source server.
+
+  * It is currently not possible to use a single ``source`` job with multiple ``pull`` jobs.
+    The way zfs abstractions are named, which can be seen in ``zrepl zfs-abstraction list``, include the name of the ``source`` job, but not the name of the client identity of the target servers (see :issue:`380`).
+    Multiple ``pull`` jobs with one ``source`` job would share the same zfs replication cursor bookmark and potentially break incremental replication.
+    Thus, each client with a ``pull`` job must have its own corresponding ``source`` job.
+
+  * With fan-out replication, the datasets referenced by each ``source`` job are likely not disjoint.
+    This can work if snapshotting and sender-side pruning is done by a separate ``snap`` job.
+    Snapshotting should be disabled in each ``source`` job on the source server and sender-side pruning disabled in the ``pull`` job on each target server.
+
+  * Note that if the datasets on the source server are restored from one of the ``pull`` targets, the replication cursor bookmarks might no longer exist on the restored system.
+    This may cause incremental replication after the restore to break.
+
+  * See :ref:`the fan-out replication quick-start guide <quickstart-fan-out-replication>` for an example of this setup.
+
 
 **Setups that do not work**:
 
--- a/docs/quickstart/fan_out_replication.rst
+++ b/docs/quickstart/fan_out_replication.rst
@@ -1,0 +1,93 @@
+.. include:: ../global.rst.inc
+
+.. _quickstart-fan-out-replication:
+
+Fan-out replication
+===================
+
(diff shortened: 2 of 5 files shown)


chenxiaolong:
This is my first time writing RST, so please let me know if I messed up the formatting somewhere. I built the docs locally with `sphinx-build docs out -c docs/sphinxconf` and it _seems_ to look okay.

docs/configuration/overview.rst
```diff
@@ -293,6 +293,10 @@ This section might be relevant to users who wish to *fan-in* (N machines replica
   * ``sink`` constrains each client to a disjoint sub-tree of the sink-side dataset hierarchy ``${root_fs}/${client_identity}``.
     Therefore, the different clients cannot interfere.
 
+* Fan-out replication: 1 ``pull`` job on each target server, each with a corresponding unique ``source`` job on the source server
+
+  * With fan-out replication, the datasets referenced by each ``source`` job are likely not disjoint. This can work if snapshotting and sender-side pruning is done by a separate ``snap`` job. Snapshotting should be disabled in each ``source`` job on the source server and sender-side pruning disabled in the ``pull`` job on each target server.
```
problame, line 298: Please use separate lines per sentence. Align with spaces to the first character after the `*`, then it works.

Additional remark on content: the reader should keep in mind that if they restore the `source` from one of the `pull` sites, the restored system will not have the replication cursor bookmarks created for the other `pull` jobs targets. Hence, it is not guaranteed that incremental replication to the other `pull` jobs will work after a restore.

docs/quickstart/fan_out_replication.rst
```diff
@@ -0,0 +1,99 @@
...
+
+Caveats
+-------
+
+Note that it is currently not possible to use a single ``source`` job with multiple ``pull`` jobs. 
+The way zfs abstractions are named, which can be seen in ``zrepl zfs-abstraction list``, include the name of the ``source`` job, but not the name of the client identity of the target servers (see :issue:`380`).
+Multiple ``pull`` jobs with one ``source`` job would share the same zfs replication cursor bookmark and potentially break incremental replication.
+Thus, each client with a ``pull`` job must have its own corresponding ``source`` job.
+
+Since there are multiple ``source`` jobs referencing the same datasets, snapshotting and pruning should be done by a separate ``snap`` job.
```
problame, line 20: Do you think this section would be better placed in `docs/configuration/overview.rst`? We should definitely link to it from here, though. Probably from within a `.. WARN:` box

docs/configuration/overview.rst
```diff
@@ -293,6 +293,10 @@ This section might be relevant to users who wish to *fan-in* (N machines replica
   * ``sink`` constrains each client to a disjoint sub-tree of the sink-side dataset hierarchy ``${root_fs}/${client_identity}``.
     Therefore, the different clients cannot interfere.
 
+* Fan-out replication: 1 ``pull`` job on each target server, each with a corresponding unique ``source`` job on the source server
```
problame, line 296: full stop at end of sentence.

docs/quickstart/fan_out_replication.rst
```diff
@@ -0,0 +1,99 @@
+.. include:: ../global.rst.inc
+
+.. _quickstart-fan-out-replication:
+
+Fan-out replication
+===================
+
+This quick-start example demonstrates how to implement a fan-out replication setup where datasets on a server (A) are replicated to multiple targets (B, C, etc.).
```
problame, line 8: There should be a reference to the section that you added to `docs/configuration/overview.rst`. Search for `:ref:` anywhere in the docs. And maybe throw in a back-reference to this quick-start example as well :)

problame (changes requested):
Thanks for this PR, here's my review.

The example config is automatically tested, and the test fails:
```
 build_jobs_test.go:141: no additional checks
    --- FAIL: TestSampleConfigsAreBuiltWithoutErrors/../../config/samples/quickstart_fan_out_replication_source.yml (0.00s)
        build_jobs_test.go:150: error parsing ../../config/samples/quickstart_fan_out_replication_source.yml:
            yaml: unmarshal errors:
              line 2: cannot unmarshal !!seq into config.Config
```

You can reproduce this locally by running `make test-go`.

docs/configuration/overview.rst
chenxiaolong: Thanks! I've added the additional remark as suggested.

chenxiaolong:
Thanks for the review! I've fixed the tests (I accidentally omitted the `jobs:` key) and updated the docs based on your comments.

Issue #553, opened 2022-01-03
Prometheus metric to track daemon start time


lapo-luchini:
In order to track upgrades/deploys (and the seldom unexpected restart) I'd like to add a "time of startup" (constant) metric.

I'd do a PR if you think it would be nice for everyone to have, I have two ideas in mind:

Right now:
```
zrepl_version_daemon{raw="v0.5.0-rc1",version_info="zrepl (…)"} 1
```

Option A (changes existing metric content, but should change nothing substantial for anyone, as current value had no meaning):
```
zrepl_version_daemon{raw="v0.5.0-rc1",version_info="zrepl (…)"} 1.6412323377629526e+09
```

Option B (add specific metric with specific value):
```
zrepl_version_daemon{raw="v0.5.0-rc1",version_info="zrepl (…)"} 1
zrepl_start_time 1.6412323377629526e+09
```

problame:
I think `Option A`, plus renaming the metric to `zrepl_uptime` is fine.
We don't provide a stability guarantee for the Prometheus metrics.
You'll have to update the Grafana dashboard as part of the PR, though :)

lapo-luchini:
Commit 67a1bbd98009b4bbcdc76ac5412aaafaaa076597 is just the bare minimum change to implement this and keep the Grafana dashboard identical as before.

But on the Grafana side it could make sense to make it retro-compatible, maybe, with a query like this? (or given the "no stability guarantee" you think this is unneeded?)
```PromQL
sgn(zrepl_uptime{job='$prom_job_name'}) or zrepl_daemon_version{job='$prom_job_name'}
```

And to improve it, adding a "uptime" panel like this:
```JSON
[...]

PR #556, opened 2022-01-12
Convert `zrepl_version_daemon` to `zrepl_uptime`.

lapo-luchini:
Closes GitHub issue #553.

### DIFF
--- a/dist/grafana/grafana-prometheus-zrepl.json
+++ b/dist/grafana/grafana-prometheus-zrepl.json
@@ -249,7 +249,7 @@
             "stack": true,
             "steppedLine": false,
             "targets": [{
-                "expr": "sgn(zrepl_uptime{job='$prom_job_name'})",
+                "expr": "sgn(zrepl_start_time{job='$prom_job_name'})",
                 "format": "time_series",
                 "interval": "",
                 "intervalFactor": 1,
--- a/dist/grafana/grafana-prometheus-zrepl.json
+++ b/dist/grafana/grafana-prometheus-zrepl.json
@@ -249,7 +249,7 @@
             "stack": true,
             "steppedLine": false,
             "targets": [{
-                "expr": "zrepl_version_daemon{job='$prom_job_name'}",
+                "expr": "sgn(zrepl_uptime{job='$prom_job_name'})",
                 "format": "time_series",
                 "interval": "",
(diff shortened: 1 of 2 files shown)


version/version.go
```diff
@@ -34,20 +34,19 @@ func (i *ZreplVersionInformation) String() string {
...
-var prometheusMetric = prometheus.NewUntypedFunc(
-	prometheus.UntypedOpts{
+var prometheusMetric = prometheus.NewGauge(
+	prometheus.GaugeOpts{
 		Namespace: "zrepl",
-		Subsystem: "version",
-		Name:      "daemon",
-		Help:      "zrepl daemon version",
+		Name:      "uptime",
+		Help:      "zrepl daemon start time and version",
```
problame, line 41: When I hear `uptime`, I expect the *duration* from the time since the thing in question started.
This metric here is not a duration but a timestamp, right?
In that case, `zrepl_start_time`?

dist/grafana/grafana-prometheus-zrepl.json
```diff
@@ -249,7 +249,7 @@
             "stack": true,
             "steppedLine": false,
             "targets": [{
-                "expr": "zrepl_version_daemon{job='$prom_job_name'}",
+                "expr": "sgn(zrepl_uptime{job='$prom_job_name'})",
```
problame, line 252: Re your question about backwards-compatible panel in Grafana (https://github.com/zrepl/zrepl/issues/553#issuecomment-1011207474)
I wouldn't bother with it, as long as we mention it in the changelog.

problame (changes requested):
See comment about term `uptime`.

version/version.go
lapo-luchini: I did suggest `start_time` as a name. 😇
(one generally wants constant values as metrics, the uptime can easily be calculated in PromQL using `time()`)

dist/grafana/grafana-prometheus-zrepl.json
lapo-luchini: Okay.

version/version.go
problame: Sorry, should have listened more carefully :)

problame:
Thanks for this contribution!

Issue #467, opened 2021-04-12
Configcheck: have a mode to ignore missing files


grahamc:
It'd be really neat if configcheck could be told to not check to see if keys or pems exist on disk. This could be used to validate configuration files for validity before they're deployed to the machines that will ultimately use them. For NixOS, this means we can validate the config before it even finishes building the system's profile.

PR #587, opened 2022-03-30
Add `--parse-only` flag to `zrepl configcheck` to prevent checking cert files

cole-h:
It may be desirable to check that a config is valid without checking for
the existence of certificate files (e.g. when validating a config inside
a sandbox without access to the cert files).

---

Closes https://github.com/zrepl/zrepl/issues/467.

It's possible this is too intrusive, but this would be very useful for NixOS so that we can check the config file at build time (e.g. potentially without proper permissions to read cert files for a TLS connection).


daemon/job/build_jobs.go
```diff
@@ -12,9 +12,17 @@ import (
...
 
 func JobsFromConfig(c *config.Config) ([]Job, error) {
+	return jobsFromConfigImpl(c, false)
+}
+
+func ParseJobsFromConfig(c *config.Config) ([]Job, error) {
+	return jobsFromConfigImpl(c, true)
+}
+
+func jobsFromConfigImpl(c *config.Config, parseOnly bool) ([]Job, error) {
```
problame, line 22: Instead of an impl method that takes a `bool`, I suggest we just add a `flags config.ParseFlags` argument to `JobsFromConfig`.
In package `config`, define a `type ParseFlags uint` that is a bitmask.
One bit in that bitmask should be `ParseFlagsNoCertCheck`.

problame (changes requested):
Thanks for this PR!

Apart from the `flags` thing in the review comment, I suggest we name the option `--skip-cert-check` or similar.

cole-h:
Sorry for the radio silence -- I'll be able to address your comments some time next week, now that finals are over.

problame:
@cole-h no worries, I've been busy myself. What is your timeline for this?

cole-h:
:facepalm: Yeah, I'll get back to this after work today. I've set a reminder, so hopefully I don't let it slip... again...

problame:
pushed a small patch on top of your branch. Does that solve your headache?

Also, can you look into whether we can get rid of

https://github.com/zrepl/zrepl/blob/95fc299733c0833ba2d12d023586fa0fb1cb69c8/transport/tls/serve_tls.go#L28-L30

using your changes?

cole-h:
I've replaced the `fakeCertificateLoading` stuff with `ParseFlagsNoCertCheck`. PTAL!

problame (approved):

problame:
@cole-h thanks for this contribution, and your patience!

PR #592, opened 2022-04-10
configurable initial replication policy

problame:
This is the finalization of work in https://github.com/zrepl/zrepl/pull/470

New syntax:
```yaml
- type: push | pull
  conflict_resolution:
    initial_replication: most_recent | all | fail
```

@grahamc @budachst would you like to give this a spin?


InsanePrawn:
I have a semantics question:
does `fail` fail a) only the dataset or b) the entire replication run?
If b), should there be a `skip`?

docs/configuration/conflict_resolution.rst
```diff
@@ -0,0 +1,35 @@
...
+
+.. _conflict_resolution-initial_replication-option-send_all_snapshots:
+
+
+``initial_replication`` option
+------------------------------
+
+The ``initial_replication`` option determines which snapshots zrepl replicates if the filesystem has not been replicated before.
+If ``most_recent`` (the default), the initial replication will only transfer the most recent snapshot, while ignoring previous snapshots.
+If all snapshots should be replicated, specify ``all``.
```
InsanePrawn, line 26: Am I correct in assuming this will still only handle snapshots matched by the job's regex? If so, we should probably mention it in this sentence.

docs/configuration/conflict_resolution.rst
```diff
@@ -0,0 +1,35 @@
...
+
+     ...
+
+.. _conflict_resolution-initial_replication-option-send_all_snapshots:
+
+
+``initial_replication`` option
+------------------------------
+
+The ``initial_replication`` option determines which snapshots zrepl replicates if the filesystem has not been replicated before.
```
InsanePrawn, line 24:
```suggestion
The ``initial_replication`` option determines how many snapshots zrepl replicates if the filesystem has not been replicated before.
```

wdyt? might be an alternative to the comment about `all` below.

problame:
Sorry for the late reply.

>  does fail fail a) only the dataset or b) the entire replication run?

Only the dataset (and any child datasets, which by definition can't exist on the receiver yet because it's an initial replication that is being failed)

docs/configuration/conflict_resolution.rst
problame: There is no such thing. Replication always sends all snapshots that it sees.

docs/configuration/conflict_resolution.rst
problame: I like that

problame:
Force-push diff: 

```
diff --git a/client/status/viewmodel/render.go b/client/status/viewmodel/render.go
index 293efc8..aaaff77 100644
--- a/client/status/viewmodel/render.go
+++ b/client/status/viewmodel/render.go
@@ -228,8 +228,16 @@ func printFilesystemStatus(t *stringbuilder.B, rep *report.FilesystemReport, max
 	}
 
 	userVisisbleCurrentStep, userVisibleTotalSteps := rep.CurrentStep, len(rep.Steps)
-	if len(rep.Steps) > 0 {
-		userVisisbleCurrentStep = rep.CurrentStep + 1 // CurrentStep is an index that starts at 0
+	// `.CurrentStep` is == len(rep.Steps) if all steps are done.
+	// Until then, it's an index into .Steps that starts at 0.
+	// For the user, we want it to start at 1.
+	if rep.CurrentStep >= len(rep.Steps) {
+		// rep.CurrentStep is what we want to show.
+		// We check for >= and not == for robustness.
+	} else {
+		// We're not done yet, so, make step count start at 1
+		// (The `.State` is included in the output, indicating we're not done yet)
+		userVisisbleCurrentStep = rep.CurrentStep + 1
 	}
 	status := fmt.Sprintf("%s (step %d/%d, %s/%s)%s",
 		strings.ToUpper(string(rep.State)),
diff --git a/docs/configuration/conflict_resolution.rst b/docs/configuration/conflict_resolution.rst
index f753a91..526e37d 100644
--- a/docs/configuration/conflict_resolution.rst
+++ b/docs/configuration/conflict_resolution.rst
@@ -21,9 +21,10 @@ Conflict Resolution Options
 ``initial_replication`` option
 ------------------------------
 
-The ``initial_replication`` option determines which snapshots zrepl replicates if the filesystem has not been replicated before.
+The ``initial_replication`` option determines how many snapshots zrepl replicates if the filesystem has not been replicated before.
 If ``most_recent`` (the default), the initial replication will only transfer the most recent snapshot, while ignoring previous snapshots.
 If all snapshots should be replicated, specify ``all``.
+Use ``fail`` to make replication of the filesystem fail in case there is no corresponding fileystem on the receiver.
 
 For example, suppose there are snapshosts ``tank@1``, ``tank@2``, ``tank@3`` on a sender.
 Then ``most_recent`` will replicate just ``@3``, but ``all`` will replicate ``@1``, ``@2``, and ``@3``.
diff --git a/replication/logic/replication_logic.go b/replication/logic/replication_logic.go
index b9dfc6e..36fedb3 100644
--- a/replication/logic/replication_logic.go
+++ b/replication/logic/replication_logic.go
@@ -238,6 +238,11 @@ func NewPlanner(secsPerState *prometheus.HistogramVec, bytesReplicated *promethe
 
 func tryAutoresolveConflict(conflict error, policy ConflictResolution) (path []*pdu.FilesystemVersion, reason error) {
 
+	if _, ok := conflict.(*ConflictMostRecentSnapshotAlreadyPresent); ok {
+		// replicatoin is a no-op
+		return nil, nil
+	}
+
 	if noCommonAncestor, ok := conflict.(*ConflictNoCommonAncestor); ok {
 		if len(noCommonAncestor.SortedReceiverVersions) == 0 {
 
@@ -494,23 +499,22 @@ func (fs *Filesystem) doPlanning(ctx context.Context) ([]*Step, error) {
 			return nil, conflict
 		}
 		if len(path) == 0 {
-			panic(fmt.Sprintf("len(path) is 0, but shouldn't be if conflict==nil: %#v", path))
-		}
-		if len(path) == 1 {
+			steps = nil
+		} else if len(path) == 1 {
 			panic(fmt.Sprintf("len(path) must be two for incremental repl, and initial repl must start with nil, got path[0]=%#v", path[0]))
-		}
-
-		steps = make([]*Step, 0, len(path)) // shadow
-		for i := 0; i < len(path)-1; i++ {
-			steps = append(steps, &Step{
-				parent:   fs,
-				sender:   fs.sender,
-				receiver: fs.receiver,
-
-				from:    path[i], // nil in case of initial repl
-				to:      path[i+1],
-				encrypt: fs.policy.EncryptedSend,
-			})
+		} else {
+			steps = make([]*Step, 0, len(path)) // shadow
+			for i := 0; i < len(path)-1; i++ {
+				steps = append(steps, &Step{
+					parent:   fs,
+					sender:   fs.sender,
+					receiver: fs.receiver,
+
+					from:    path[i], // nil in case of initial repl
+					to:      path[i+1],
+					encrypt: fs.policy.EncryptedSend,
+				})
+			}
 		}
 	}
 
```

InsanePrawn:
I haven't attempted to review the actual code, but from the docs this LGTM. :+1: 

> > does `fail` fail a) only the dataset or b) the entire replication run?
> 
> Only the dataset (and any child datasets, which by definition can't exist on the receiver yet because it's an initial replication that is being failed)

I think that assumption can be broken when you change a dataset from excluded to included while its children were always included - but seems fair enough to me for now.
That edge case should be handled when we tackle conflict resolution for 'placeholder exists' or the more general 'no common ancestor', either way not in this MR.

PR #596, opened 2022-04-24
Change to 2 or more machines

JMoVS:


docs/configuration/overview.rst
```diff
@@ -245,7 +245,7 @@ Limitations
 
 .. _jobs-multiple-jobs:
 
-Multiple Jobs & More than 2 Machines
+Multiple Jobs & 2 or more Machines
```
problame, line 248: What was it again that you found confusing here?

The section covers anything related to setups with more than one sender-receiver job pair.

JMoVS: yes, but in my case, I needed multiple jobs, had 2 machines but not more than 2 machines and yet was in this section. In the case of a pulling from a source that is not connected often, one can need multiple jobs and yet have exactly 2 machines. Depending on how you interpret the "&" (is it an OR or an AND), it is a bit clearer to specify "2 or more machines"

problame: Yeah, so, I think the meaning was (and still is) OR.
Basically, the section tells you to be cautious if more than 2 jobs or more than 2 machines are involved in your zrepl setup.

JMoVS: no, in my case I'm affected by this section and I have only 2 machines, not more than 2 machines.

JMoVS: so it could be "more than 1 machine" or "2 or more machines", but "more than 2 machines" suggests at leats 3, which is not my case yet I needed to look into this section :)

problame: You have more than two jobs.

JMoVS: yes, I have more than 2 jobs but onyl 2 machines, not more than 2 machine :)

Issue #600, opened 2022-05-05
Feature request: Different retention times for different filesystems (via fs specific prefixes or fs specific prune rules)


callumgare:
I currently have serverA setup to push a number of filesystems to serverB. Some of those filesystems have data that I'd like to keep snapshots of for a long time, some I only need to keep for a short time. It would be great to have an easy way to specify that. There's two different approaches that occur to me that could work.

You could enable this by allowing the snapshot prefix to be different for different filesystems. That way the existing pruning options could be used for also filtering by filesystems. One advantage of this approach is that since it's a more general solution it could be useful for other use-cases as well (although I can't think of any of the top of my head). An example of this might look like:

```yaml
jobs:
  - name: push-to-serverB
    type: push
    connect:
      ...
    filesystems: {
      "data/very-important": true
      "data/not-so-important" true
    }
    snapshotting:
      type: periodic
      interval: 10m
      prefix-per-filesystem: {
        "data/very-important":  zrepl_important_
        "data/not-so-important":  zrepl_not-as-important_
      }
    pruning:
      keep_sender:
        - type: not_replicated
        - type: grid
          grid: 1x1h(keep=all) | 24x1h | 5x1d
          regex: "^zrepl_not-as-important_.*"
        - type: grid
          grid: 1x1h(keep=all) | 24x1h | 5x1d |  6x30d
          regex: "^zrepl_important_.*"
      keep_receiver:
        - type: grid
          grid: 1x1h(keep=all) | 24x1h | 14x1d
          regex: "^zrepl_not-as-important_.*"
        - type: grid
          grid: 1x1h(keep=all) | 24x1h | 14x1d | 3x30d
          regex: "^zrepl_important_.*"
```

Alternatively the other option I can think of would be to add a filesystem filter to the pruning rules. For example:

```yaml
jobs:
  - name: push-to-serverB
    type: push
    connect:
      ...
    filesystems: {
      "data/very-important": true
      "data/not-so-important" true
    }
    snapshotting:
      type: periodic
      interval: 10m
      prefix: zrepl_
    pruning:
      keep_sender:
        - type: not_replicated
        - type: grid
          grid: 1x1h(keep=all) | 24x1h | 5x1d
          regex: "^zrepl_not-as-important_.*"
          filesystems: {
            "data/not-so-important" true
          }
        - type: grid
          grid: 1x1h(keep=all) | 24x1h | 5x1d |  6x30d
          regex: "^zrepl_important_.*"
          filesystems: {
[...]

PR #602, opened 2022-05-18
Add FSFilter to pruning rules

3nprob:
This adds [`filesystems` filters](https://zrepl.github.io/configuration/filter_syntax.html) support to pruning policies.
It adds an optional field `filesystems` to all pruning policies (default `{ "<": true }`).

This is intended to solve for #600

### DIFF
--- a/daemon/pruner/pruner.go
+++ b/daemon/pruner/pruner.go
@@ -98,7 +98,7 @@ func NewLocalPrunerFactory(in config.PruningLocal, promPruneSecs *prometheus.His
 		return nil, errors.Wrap(err, "cannot build pruning rules")
 	}
 	for _, r := range in.Keep {
-		if _, ok := r.Ret.(*config.PruneKeepNotReplicated); ok {
+		if _, ok := r.Ret.(*config.KeepNotReplicated); ok {
 			// rule NotReplicated  for a local pruner doesn't make sense
 			// because no replication happens with that job type
 			return nil, fmt.Errorf("single-site pruner cannot support `not_replicated` keep rule")
@@ -125,7 +125,7 @@ func NewPrunerFactory(in config.PruningSenderReceiver, promPruneSecs *prometheus
 
 	considerSnapAtCursorReplicated := false
 	for _, r := range in.KeepSender {
-		knr, ok := r.Ret.(*config.PruneKeepNotReplicated)
+		knr, ok := r.Ret.(*config.KeepNotReplicated)
 		if !ok {
 			continue
 		}
--- a/daemon/pruner/pruner.go
+++ b/daemon/pruner/pruner.go
@@ -98,7 +98,7 @@ func NewLocalPrunerFactory(in config.PruningLocal, promPruneSecs *prometheus.His
 		return nil, errors.Wrap(err, "cannot build pruning rules")
 	}
 	for _, r := range in.Keep {
-		if _, ok := r.Ret.(*config.KeepNotReplicated); ok {
+		if _, ok := r.Ret.(*config.PruneKeepNotReplicated); ok {
 			// rule NotReplicated  for a local pruner doesn't make sense
 			// because no replication happens with that job type
 			return nil, fmt.Errorf("single-site pruner cannot support `not_replicated` keep rule")
@@ -125,7 +125,7 @@ func NewPrunerFactory(in config.PruningSenderReceiver, promPruneSecs *prometheus
 
 	considerSnapAtCursorReplicated := false
 	for _, r := range in.KeepSender {
-		knr, ok := r.Ret.(*config.KeepNotReplicated)
+		knr, ok := r.Ret.(*config.PruneKeepNotReplicated)
 		if !ok {
 			continue
 		}
--- a/daemon/pruner/pruner.go
+++ b/daemon/pruner/pruner.go
@@ -291,6 +291,7 @@ const (
 	NotSkipped                   = ""
 	SkipPlaceholder              = "filesystem is placeholder"
 	SkipNoCorrespondenceOnSender = "filesystem has no correspondence on sender"
+	SkipNoRules                  = "filesystem did not match any rules"
 )
 
 func (r FSSkipReason) NotSkipped() bool {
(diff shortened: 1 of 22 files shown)


docs/configuration/prune.rst
```diff
@@ -219,6 +219,33 @@ Policy ``regex``
 Like all other regular expression fields in prune policies, zrepl uses Go's `regexp.Regexp <https://golang.org/pkg/regexp/#Compile>`_ Perl-compatible regular expressions (`Syntax <https://golang.org/pkg/regexp/syntax>`_).
 The optional `negate` boolean field inverts the semantics: Use it if you want to keep all snapshots that *do not* match the given regex.
 
+.. _prune-filters:
+
+Filesystem filters in pruning policies
```
problame, line 224: Title case: capitalize each word's first character

docs/configuration/prune.rst
```diff
@@ -219,6 +219,33 @@ Policy ``regex``
...
+
+::
+
+    jobs:
+    - type: pull
+      pruning:
+        keep_receiver:
+        - type: grid
+          filesystems:
+            "tank/long": true
```
problame, line 238: Are these filesystem names sender-side or receiver side?

Regardless, please provide the `root_fs` of this `pull` job in the example to clarify.

problame (changes requested):

daemon/pruner/pruner.go
```diff
@@ -392,12 +394,51 @@ tfss_loop:
...
+				return
+			}
+			if match {
+				rules = append(rules, r)
+			}
+		}
+
+		if len(rules) == 0 {
+			pfs.skipReason = SkipNoRules
+			l.WithField("skip_reason", pfs.skipReason).WithField("sfs", sfs.GetPath()).Debug("skipping filesystem")
```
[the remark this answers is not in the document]
3nprob: Added in changelog. I figured a mistake is better resulting in retention that accidental destruction

docs/configuration/prune.rst
3nprob: Hope I got it right - didn't include an explicit `root_fs` but mirrored my tested config with a similar format to existing docs

problame:
Adding this for 0.6, I want to take it but haven't found time to review yet. Maybe later today

3nprob:
force-push was just rebasing with master and resolving merge conflict in docs

problame (changes requested):
Can we put `Regex` and `Filesystems` into a `KeepRuleCommon` struct that is embedded within each concrete keep rule?
And mirror that in the `config` package.

That way, it will look like

```
package config

type KeepRuleCommon struct {
  Regex
  Filesystems
}

type KeepRuleGrid struct {
  KeepRuleCommon
  GridSpec string
}
```

```
package pruning

struct keepCommon type {
  Regex re.Regexp
  Filesystems DatasetFilter
}

func newKeepCommon(in *config.KeepRuleCommon) keepCommon (

}

func NewKeepGrid(in *config.KeepRuleGrid) *KeepGrid {
  g := &KeepGrid{}
  g.keepCommon = newKeepCommon(in.KeepRuleCommon)
...
}

**Comment by 3nprob** [comment]
Sure thing, keeping in mind that `NotReplicated` has filesystems but no regex (unless you're OK with adding Regex to it?) so will still have the explicit field

**Comment by problame** [comment]
>  has filesystems but no regex (unless you're OK with adding Regex to it?)

I'm ok with adding it.

**Comment by problame** [comment]
Hm, actually, can we just check in the `KeepNotReplicated` constructor function that the regex is nil?

**Comment by 3nprob** [comment]
> Hm, actually, can we just check in the `KeepNotReplicated` constructor function that the regex is nil?

Do you mean that you'd like to see all the prunerule constructors take the config struct as sole argument, like Grid is today?

**Comment by 3nprob** [comment]
Ended up going along with that, also breaking out the `MustKeep*` functions to clarify that they're used as test helpers and not part of the application flow, and addressing an outstanding FIXME. The behavior should be the same as before.

Hopefully you agree that this makes this part of the code more consistent and clear.

Some commits will be squashed but other than that this should be ready for review now.

**Comment by 3nprob** [comment]
Just pushed an optional 195637b987d9c51439ab2af1ca0f6897db85c79e.

I don't see why the other rules that support regex should not also be able to support negation. It's not yet exposed in the config though, as that's a separate issue. Just figured I might as well accommodate for it.

**Comment by problame** [comment]
Pushed some easy bulk edits and some docs improvements on top of your branch.
Mostly to make the diff smaller.
We should have them, but in a later commit.
Will review now.

File: pruning/keep_grid.go [hunk 4]
```diff
@@ -15,42 +14,31 @@ import (
...
-	if err != nil {
-		panic(err)
-	}
-	return grid
-}
-
-func newKeepGrid(re *regexp.Regexp, configIntervals []config.RetentionInterval) (*KeepGrid, error) {
-	if re == nil {
-		panic("re must not be nil")
+	if common.filesystems == nil {
```
problame, line 40: Why do we have this test for `keep_grid.go`, but not the other keep rules?
Shouldn't `newKeepCommon`, in combination with the default value in `config`, guarantee that we always have a non-nil filesystems filte?

pruning/keep_grid.go
```diff
@@ -82,16 +70,20 @@ func newKeepGrid(re *regexp.Regexp, configIntervals []config.RetentionInterval)
...
+		retentiongrid.NewGrid(intervals),
 	}, nil
 }
 
 // Prune filters snapshots with the retention grid.
 func (p *KeepGrid) KeepRule(snaps []Snapshot) (destroyList []Snapshot) {
 
 	matching, notMatching := partitionSnapList(snaps, func(snapshot Snapshot) bool {
-		return p.re.MatchString(snapshot.Name())
+		if p.negate {
```
problame, line 82: Please extract the `if p.negate  { ... } else { ... }` logic into `KeepCommon`. I see it repeated in every `KeepRule` implementation.
You can use a private (`lower-caps`) method on `KeepCommon` for that.

pruning/pruning.go
```diff
@@ -2,15 +2,29 @@ package pruning
...
+	"github.com/zrepl/zrepl/daemon/filters"
+	"github.com/zrepl/zrepl/zfs"
 )
 
 type KeepRule interface {
 	KeepRule(snaps []Snapshot) (destroyList []Snapshot)
+	GetFSFilter() zfs.DatasetFilter
+}
+
+type KeepCommon struct {
```
problame, line 20: Does this need to be an exported struct? I don't think so.

pruning/pruning_test.go
```diff
@@ -83,7 +83,7 @@ func TestPruneSnapshots(t *testing.T) {
 		"simple": {
 			inputs: inputs["s1"],
 			rules: []KeepRule{
-				MustKeepRegex("foo_", false),
+				MustKeepRegex(map[string]bool{}, "foo_", false),
```
problame, line 86: Please find a way to avoid the clutter with the empty map this test function.
And please add another test function that tests the functionality which you added.

To accomplish the former, I suggest the following steps:
1. you introduce the `KeepCommon` concept to the MustKeep* family of functions.
2. use a default `KeepCommon` for all MustKeep calls within this function. That default doesn't have a `filesystems` filter.

Then, for the additional tests, you'll likely just vary the KeepCommon.

zfs/zfs.go
```diff
@@ -1919,3 +1919,21 @@ func ZFSRollback(ctx context.Context, fs *DatasetPath, snapshot FilesystemVersio
 
 	return err
 }
+
+func FilterCheckFS(fsfilter DatasetFilter, fs string) (*DatasetPath, error) {
```
problame, line 1923: Can we have this in a separate PR? Maybe we can start using generics for this.

problame (changes requested):
Making good progress, keep up the good work!

pruning/keep_grid.go
3nprob: You're right, good catch!

zfs/zfs.go
3nprob: Do you mean just break this one hunk into a separate PR, which the current one then gets dependent on? (As the meat of  this PR is inherently coupled to this addition)

If so; sure, if not; I may need some more elaboration on what's intended

problame: > As the meat of this PR is inherently coupled to this addition

Why is that?

3nprob: ...I'll get some sleep and get back to this :face_in_clouds:

problame:
@3nprob are you still pursuing this effort?

problame:
Closing this due to inactivity.

PR #604, opened 2022-05-19
Allow setting destroy max batch size

### LINKED ISSUE #508: pruning: option to specify maximum batch size
> In my experience, depending on the system and amount of data being held by the snapshots, deleting even 100 snapshots at a time might overwhelm the system. If you do implement this, please allow the number of snapshots deleted at once to be configurable (as either number of snapshots or "size" of snapshots).
> Thank you!

_Originally posted by @helamonster in https://github.com/zrepl/zrepl/issues/72#issuecomment-896386446_

3nprob:
Add env var `ZREPL_DESTROY_MAX_BATCH_SIZE` to allow changing the max batch size for batch destroys. The default is set to 200.

Resolves #508

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,6 +21,7 @@ Developers should consult the git commit log or GitHub issue tracker.
 
 * `Feature Wishlist on GitHub <https://github.com/zrepl/zrepl/discussions/547>`_
 
+* |feature| Add ``ZREPL_DESTROY_MAX_BATCH_SIZE`` env var (default 0=unlimited).
 * |break| |feature| convert Prometheus metric ``zrepl_version_daemon`` to ``zrepl_start_time`` metric
 
   * The metric still reports the zrepl version in a label.
@@ -275,7 +276,7 @@ Changes
 * |feature| Proper timeout handling for the :ref:`SSH transport <transport-ssh+stdinserver>`
 
   * |break| Requires Go 1.11 or later.
-  
+
 * |break| |break_config|: mappings are no longer supported
 
   * Receiving sides (``pull`` and ``sink`` job) specify a single ``root_fs``.
--- a/zfs/versions_destroy.go
+++ b/zfs/versions_destroy.go
@@ -115,9 +115,10 @@ func buildBatches(reqs []*DestroySnapOp) [][]*DestroySnapOp {
 	// group by fs
 	var perFS [][]*DestroySnapOp
 	consumed := 0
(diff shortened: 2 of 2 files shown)


problame:
Thanks for your contribution. A few questions:
* Is there any specific reason why you chose 200 as the default?
* Would you be willing to put in the effort to make this a yml config variable?

3nprob:
@problame 
> Is there any specific reason why you chose 200 as the default? 
Mostly finger + air but basically ran into memory exhaustion on 300 on a 4GB system so seemed like a reasonable compromise given little noise about this from others.
But I can see down to 50 or so being reasonable to have a safe default for resource-constrained environments.

> Would you be willing to put in the effort to make this a yml config variable?

I actually started with that but ended up abandoning it - do you think you could provide some guidance of where to inject the config? I traced the calls until I got to e.g. https://github.com/zrepl/zrepl/blob/1acafabb5bd5355c838403465a554c1a1134ce1d/endpoint/endpoint.go#L398, where now we reach the point of sending it over protobuf (and I figured this is the reason much of other global config is also only exposed via env vars)...
Would it make sense to change the intialization of `destroyerSingleton` and have it there?
But maybe there is a neater way to reach to it that I didn't realize.

(FWIW I also figured a "dry run" option for pruning would make sense and that's pretty much the same considerations as here - which is just to say it would be good for any solution to not introduce too much tech debt if other config with similar scope are introduced)

problame:
> , where now we reach the point of sending it over protobuf (and I figured this is the reason much of other global config is also only exposed via env vars).

It would need to be somewhere in the `global` section, I think. 

My problems with this PR:
- envconst's shouldn't be documented
- the default should remain unlimited, in my opinion. It's ZFS's job to avoid OOMs.

Regarding the latter point:

> Mostly finger + air but basically ran into memory exhaustion somewhere around 300 on a 4GB system so seemed like a reasonable compromise given little noise about this from others.

Was the memory exhaustion due to ZFS or did zrepl allocate too much memory?

3nprob:
> > , where now we reach the point of sending it over protobuf (and I figured this is the reason much of other global config is also only exposed via env vars).
> 
> It would need to be somewhere in the `global` section, I think.
> 
> My problems with this PR:
> 
>     * envconst's shouldn't be documented
> 
>     * the default should remain unlimited, in my opinion. It's ZFS's job to avoid OOMs.
> 

Noted, will try to address this.


> Was the memory exhaustion due to ZFS or did zrepl allocate too much memory?

ZFS.  Manually running the same `zfs destroy` with the same arguments also causes deadlock.

PR #613, opened 2022-07-10
revise `encrypted` send option & resume token validation of it

problame:
From the changelog

```
* |bugfix| Fix resuming from interrupted replications that use ``send.raw`` on unencrypted datasets.

  * The send options introduced in zrepl 0.4 allow users to specify additional zfs send flags for zrepl to use.
    Before this fix, when setting ``send.raw=true`` on a job that replicates unencrypted datasets,
    zrepl would not allow an interrupted replication to resume.
    The reason were overly cautious checks to support the ``send.encrypted`` option.
  * This bugfix removes these checks from the replication planner.
    This makes ``send.encrypted`` a sender-side-only concern, much like all other ``send.*`` flags.
  * However, this means that the ``zrepl status`` UI no longer indicates whether a replication step uses encrypted sends or not.
    The setting is still effective though.
```


problame:
@InsanePrawn @infisil any chance you can take a look / test drive this? For my setup (which doesn't use unencrypted raw sends) I can confirm there have been no crashes / regressions  in the last 21 days.

problame:
@infinisil , sorry for misspelling

Issue #554, opened 2022-01-04
Include possibility for cron snapshotting


khenderick:
Currently only `manual` and `periodic` are supported in the `snapshotting` > `type` section. I think it might make sense to for example introduce the type `cron`. In a new `cron` key, a user could configure something like `0 10 * * *` to take a snapshot every day at 10:00, and afterwards e.g. replicate it to another node.

Example:
```
jobs:
- type: push
  filesystems: {
    "<": true,
    "tmp": false
  }
  snapshotting:
    type: cron
    cron: 0 10 * * *
    prefix: zrepl_
    hooks: ...
```

The use case for which I would like to use this is to use a hook to take a backup from a non-zfs machine and download it to a local dataset. After the hook completes, the dataset is snapshotted and replicated. I want this to run on a specific time to make sure the backup has minimal impact on the non-zfs machine.

This ticket is similar to #397 but is not focused on replication only

problame:
Thanks for opening this issue.
I think it's a good issue to get started with zrepl development.
Would you like to draw attention to this [in the discussion for zrepl 0.6](https://github.com/zrepl/zrepl/discussions/547)?

problame:
I'm working on this

Issue #394, opened 2020-10-12
Request: ability to run pre/post hooks on a push job


DavieDavieDave:
I have a remote machine configured as a 'sink' to an encrypted dataset. I'd like this dataset to unmounted and locked when the data is at rest. At the moment, I have a cron job that mounts and unlocks the dataset remotely prior to waking up the push job. What I would like is the ability to run pre and post command prior to/after running the job. I see something similar in issue #74 for snapshots.

problame:
This feature requests and reasonable and should be fairly easy to implement for beginner. A few questions:

What metadata will you need in the hook script?
What should happen if the hook times out or exits with non-zero status code?

Also: I suppose your use case is sending from a pool that doesn't support OpenZFS encryption into one that does?

problame:
(To get an impression of the metadata that's available on the post-edge, run `zrepl status --raw` after a job completed its invocation).

DavieDavieDave:
> What metadata will you need in the hook script?

For my particular use case, I don't need any metadata at this time, but I could see some of it being useful in the future. Maybe providing items such as the job state and errors of the job would allow a script to have better error handling.

> What should happen if the hook times out or exits with non-zero status code?

In my scenario, I would expect the job to fail as the encrypted dataset on the remote side would not be reachable or the dataset is not mounted.

> Also: I suppose your use case is sending from a pool that doesn't support OpenZFS encryption into one that does?

Correct.

> (To get an impression of the metadata that's available on the post-edge, run `zrepl status --raw` after a job completed its invocation).

```
[...]

PR #614, opened 2022-07-10
refactor snapper & support cron-based snapshotting

problame:
fixes https://github.com/zrepl/zrepl/issues/554
refs https://github.com/zrepl/zrepl/discussions/547#discussioncomment-1936126


problame:
@khenderick @avg-l @MarcusWichelmann @spi43984 @janprzy
You have been involved in this feature request, either via the [0.6 Wishlist](https://github.com/zrepl/zrepl/discussions/547#discussioncomment-1936126) or the issue https://github.com/zrepl/zrepl/issues/554

Please give this PR a test drive and report whether you find anything lacking.
I'd like to get it merged soon, as this PR touches the hooks code, and there is some other interest in hooks https://github.com/zrepl/zrepl/issues/394

problame:
> Given a cron of 0 21 * * *. Assume the server is offline at 21:00 but comes back online at 21:30. Will zrepl immediately take the snapshot it missed, or will it wait for the next day at 21:00?

The implementation is in `cron.go` in the `Run` function. Basically, it computes the wake time using the cron library (line 54), then programs a Go runtime timer for that wake time, and goes to sleep until either that timer fires or the the daemon is stopped.
From there on, rinse and repeat.

So, if `offline` means suspend/hibernate and resume, the kernel will fire the timer and zrepl will take the snapshot delayed.
I guess I need to fix that aspect, i.e., when the timer fires, check current wallclock time again and if we're too far off from originally targeted wall clock time, don't take a snapshot. That's what cron does when you suspend&resume afaik.

But if `offline` means daemon stopped (or even power out), then zrepl will compute a new timer when it's restarted. But it won't take the last time snapshots were taken into consideration.

> I noticed support for @daily, which as far as I understand means 0 0 * * *. If the answer on the first question is that it will wait for the next moment, I assume it will be the same for @daily. In that case it might be interesting to give @daily a slightly different interpretation where it means "at the first possible moment", which would be 00:00 if the server would be running, but could be e.g. 03:27 if that was when zrepl started after the server was started up again after a maintenance window. This could then also be used to e.g. snapshot a laptop as soon as it boots.

I'm simply passing through the `cron` library here, so, think of @daily as an alias for 0 0 * * *.

I have no intention to implement any "on first boot" behavior".
That resembles systemd's `OnBootSec=`, and there's no way zrepl is going to implement such a thing reliably, since it's not `init`.
In general, I think the laptop use case is best served by the current `snapshotting.type=periodic`, or an external snapshotting mechanism.

I guess we can have another `snapshotting.type` in the future where zrepl doesn't take snaps automatically, but you can tell it to take the snapshots via the CLI.
That requires some reworking of the current `signal` mechanism though, which I have on my to-do list, but won't happen in the near future.

---

Do the limitations outlined above make this PR less appealing / useful to you?

khenderick:
> Do the limitations outlined above make this PR less appealing / useful to you?
No, my initial request for a cron-like feature was with a server in mind that can be assumed to be up-and-running at the scheduled time. So far, it seems to be working perfectly. There are indeed plenty of other options to tackle the laptop use-case.

I really appreciate your effort/work of making all of this available to us, thanks for that!

problame:
@khenderick you're welcome, thanks for testing!

> In Linux there is some collaboration between cron and ancron. Anacron makes sure a @daily cronjob is run on a daily basis - even if the system itself (i. e. the one with the cronjob) is offline.  So @daily doesn't necessarily mean 0 0 * * * but something like run it a day after the last run, if the system is offline, try it later.

Thanks @spi43984 , I always asked myself how @daily works. 
So, I think if someone is happy with that definition of `daily`, they should be happy with `periodic` set to `24h` as well.


>    For time reasons - how do I only compile for Linux/amd64 only? Currently I get arm and freebsd binaries compiled as well... I commented out the make directives in Makefile but there's probably a nicer way...
>    How do I explicitly compile the cron-snap version? Currently am getting the v0.5.0-17-g02b2151 version compiled...

1. Checkout the PR. 
2. `make`
3. Replace your installed zrepl binaries with the one from `./artifacts/zrepl-linux-amd64`

If you want a deb or rpm, just throw in an additional step

4. `make deb` or `make rpm` , or `make deb-docker` / `make rpm-docker` respectively.

spi43984:
> > For time reasons - how do I only compile for Linux/amd64 only? Currently I get arm and freebsd binaries compiled as well... I commented out the make directives in Makefile but there's probably a nicer way...
> > How do I explicitly compile the cron-snap version? Currently am getting the v0.5.0-17-g02b2151 version compiled...
> 
>     1. Checkout the PR.
> 
>     2. `make`
> 

Thanks - does that mean that docker driven compilation is not supported for the PR?

>     3. Replace your installed zrepl binaries with the one from `./artifacts/zrepl-linux-amd64`
> 
> 
> If you want a deb or rpm, just throw in an additional step
> 
>     4. `make deb` or `make rpm` , or `make deb-docker` / `make rpm-docker` respectively.

spi43984:
Compiled and tested. What I do experience is that `cron: "*/5 * * * *"` works but `cron: "32,33 17 * * *"` not.

The latter should run at 17:32 and 17:33 - which obviously doesn't make sense in this case. But something like `cron: "3 7,19 * * *" might make sense and runs a snapshot at 7:03 and 19:03. As I didn't want to wait an hour I just checked whit multiple minutes...

To get an idea what combinations might be possible, here's a link for a [cron generator](https://crontab-generator.org/)

I assume `cron: "* */12 * * *"` to run every 12 hours works as well but in that case I could have used `type: periodic` as well - so no need for cron.

If interpreting `cron: "3 7,19 * * *"` is an issue maybe multiple cron-lines within one job definition would help, like
```
cron: "3 7 * * *"
cron: "3 19 * * *"
```
or
```
cron: "3 7 * * *", "3 19 * * *"
```

problame:
@spi43984  this is the lib I use to parse the cron syntax: https://pkg.go.dev/github.com/robfig/cron/v3 .
Specifically, the `cron.ParseStandard` function: https://pkg.go.dev/github.com/robfig/cron/v3#ParseStandard

spi43984:
> @spi43984 this is the lib I use to parse the cron syntax: https://pkg.go.dev/github.com/robfig/cron/v3 . Specifically, the `cron.ParseStandard` function: https://pkg.go.dev/github.com/robfig/cron/v3#ParseStandard

My cron config looked like that `cron: "53 17,18,19 * * *"` and it only run once at 17:53 but no later. From a quick view from the cron library it supports `,`. Might there be some parsing issue with the surrounding `"` and the `,` in between?

problame:
Most certainly not, it's straight YAML parsing. My guess is that it's a bug in the library, or I'm using it wrong.

So, IIUC, it currently silently drops everything after the `,`?

problame:
@spi43984  found and fixed the issue; it was in how I used the timer API :roll_eyes: 
Please pull this branch and continue testing.

spi43984:
> @spi43984 found and fixed the issue; it was in how I used the timer API roll_eyes Please pull this branch and continue testing.

`cron: "7-9 10,11,12 * * *"` works for the minutes. About the hour part we'll see in a couple of hours but it looks very promising. Thanks for fixing!

spi43984:
> > > I noticed support for @daily, which as far as I understand means 0 0 * * *. If the answer on the first question is that it will wait for the next moment, I assume it will be the same for @daily. In that case it might be interesting to give @daily a slightly different interpretation where it means "at the first possible moment", which would be 00:00 if the server would be running, but could be e.g. 03:27 if that was when zrepl started after the server was started up again after a maintenance window. This could then also be used to e.g. snapshot a laptop as soon as it boots.
> > 
> > 
> > I'm simply passing through the `cron` library here, so, think of @daily as an alias for 0 0 * * *.
> 
> In Linux there is some collaboration between cron and ancron. Anacron makes sure a @daily cronjob is run on a daily basis - even if the system itself (i. e. the one with the cronjob) is offline. So @daily doesn't necessarily mean 0 0 * * * but something like run it a day after the last run, if the system is offline, try it later.

The cron lib doc says @daily means midnight ("0 0 * * *") - so it's different then anacron/cron in Linux.

spi43984:
> `cron: "7-9 10,11,12 * * *"` works for the minutes. About the hour part we'll see in a couple of hours but it looks very promising. 

Job got executed at 10:07, 10:08, 10:09 and again at 11:07, 11:08, 11:09, 12:07, 12:08, 12:09. Works as expected. Next thing would be to check DoW, DoM and months...

problame:
> The cron lib doc says @daily means midnight ("0 0 * * *") - so it's different then anacron/cron in Linux.

I don't think these @ "descriptors" are very useful without the integration with anacron / systemd-timerd

So, I think I'll disable them.

>  Next thing would be to check DoW, DoM and months...

I think the chances that this works are pretty good. Not going to wait that out before merging :P

problame:
One thing that came to mind though:

People might want to use `cron` to implement a snapshot management scheme where they have separate "trains" of snapshots. Say, one cron that runs at the top of the hour, creating snaps named `zrepl_hourly_...`. And another one that runs daily, creating snaps named `zrepl_daily_...`. Then, they'd have matching `last_n` type keep rules for pruning to retain, say 24 snaps that have regex matching `^zrepl_hourly` and 7 snaps that have regex matching `^zrepl_daily_` 

So, we should probably support multiple cron rules in one `snapshotting` rule. Will make those changes.

spi43984:
> > The cron lib doc says @daily means midnight ("0 0 * * *") - so it's different then anacron/cron in Linux.
> 
> I don't think these @ "descriptors" are very useful without the integration with anacron / systemd-timerd
> 
> So, I think I'll disable them.
> 
Yes, I don't see a benefit either.

> > Next thing would be to check DoW, DoM and months...
> 
> I think the chances that this works are pretty good. Not going to wait that out before merging :P

My thought as well...

spi43984:
> The last commits implements multiple cron schedules in one job. Try it with
> 
> ```
>     - type: snap
>       name: snapper
>       filesystems: {
>         "p1": true
>       }
>       snapshotting:
>         type: cron
>         schedules:
>         - cron: "1-59 * * * * *"
>           prefix: zrepl_cron_minutely_
>         - cron: "0 1-23 * * * *"
>           prefix: zrepl_cron_hourly_
>         - cron: "0 0 * * * *"
>           prefix: zrepl_cron_daily_
>       pruning:
>         keep:
>           - type: last_n
>             count: 60
>             regex: "^zrepl_cron_minutely_"
>           - type: last_n
>             count: 24
>             regex: "^zrepl_cron_hourly_"
>           - type: last_n
>             count: 7
>             regex: "^zrepl_cron_daily_"
> ```
> 
> It's useful, but not very pretty. We could do better if `snapshotting` and `pruning` were one thing called `snaphot_management`. What do you think? cc @InsanePrawn

Nice, will check. You mean something like
```
      snapshotting:
        type: cron
        schedules:
        - cron: "1-59 * * * * *"
          prefix: ^zrepl_cron_minutely_
          keep: last 60
        - cron: "0 1-23 * * * *"
          prefix: ^zrepl_cron_hourly_
          keep: last 24
        - cron: "0 0 * * * *"
          prefix: ^zrepl_cron_daily_
          keep: last 7
```

problame:
Yeah, something along those lines. But I don't want to let this scope-creep too much. I imagine what I posted in https://github.com/zrepl/zrepl/pull/614#issuecomment-1190692441 would be a net improvement for many people who have trouble with `grid`.

problame:
One other things where I'm not sure what the best thing to do is:

If two schedules fire at the same time, should we serialize the snapshotting, have it run in parallel, or not snapshot at all? The current impl is serializing because
1. ZFS serializes two concurrent `zfs snapshot` commands internally anyways and
2. If there are hooks defined, then it's guaranteed that for a given filesystem, it's always `prehook, snap, posthook, prehook, snap, posthook`, as opposed to `prehook, prehook, snap, posthook, snap, posthook` or some other interleaving.

spi43984:
What's the downsize or impact of serializing snapshots?

From a backup perspective I don't see a need for two or more concurrent snapshot events but it might happen with cron. If it's too much load for instance one could prioritise the more infrequent cron event and omit the other.

problame:
> What's the downsize or impact of serializing snapshots?

None, really, I think.

problame:
I'm thinking of backing out the multi-schedules change though. The same can be achieved with multiple `snap` jobs:

```
- type: snap
  name: snapper
  filesystems: {
    "p1": true
  }
  snapshotting:
    type: cron
    schedule:  "1-59 * * * * *"
    prefix: zrepl_cron_minutely_
  pruning:
    keep:
      - type: last_n
        count: 60
        regex: "^zrepl_cron_minutely_"
      - type: regex
        negate: true
        regex: "^zrepl_cron_minutely_"

- type: snap
  name: snapper
  filesystems: {
    "p1": true
  }
  snapshotting:
    type: cron
    schedule:  "0 1-23 * * * *"
    prefix: zrepl_cron_hourly_
  pruning:
    keep:
      - type: last_n
        count: 24
        regex: "^zrepl_cron_hourly_"
      - type: regex
        negate: true
        regex: "^zrepl_cron_hourly_"

- type: snap
  name: snapper
  filesystems: {
    "p1": true
  }
  snapshotting:
    type: cron
    schedule: "0 0 * * * *"
    prefix: zrepl_cron_daily_
  pruning:
    keep:
      - type: last_n
        count: 7
        regex: "^zrepl_cron_daily_"
      - type: regex
        negate: true
        regex: "^zrepl_cron_daily_"
```

If it's for use with a `push` job, there'll be a need for an outside cron job / systemd timer to wake up the push job, much like we have it [in the external-disk quickstart guide](https://zrepl.github.io/quickstart/backup_to_external_disk.html).

I want to address intra-job wakeups soon, but not in the 0.6 release.
But I want the `cron` feature request to be addressed in 0.6.
So, if we can make people happy with the above, instead of the multiple-schedules thing proposed in https://github.com/zrepl/zrepl/pull/614#issuecomment-1190692441 I'd want to keep it simple.

problame:
@spi43984  @khenderick  please comment on whether the simple variant with only one schedule would be sufficient for you, given the considerations above.

spi43984:
> @spi43984 @khenderick please comment on whether the simple variant with only one schedule would be sufficient for you, given the considerations above.

The simple variant is good enough for me. Thanks for the changes!

problame:
Getting back to this now. I just force-pushed a rebase on latest `master` and reverted the multi-schedule change.
I think it's ready for merge. @spi43984  do you want to test it again?

spi43984:
> Getting back to this now. I just force-pushed a rebase on latest `master` and reverted the multi-schedule change. I think it's ready for merge. @spi43984 do you want to test it again?

Will do... Thx!

problame:
@spi43984 any updates?

spi43984:
> @spi43984 any updates?

Got knocked out by Covid for two weeks. Will try to test next coming days...

spi43984:
> @spi43984 any updates?

ok. Could run the test right now. It works. Tested two cron jobs with following config - one runs every odd minute, the other one every even one:

```
jobs:
- name: TLS1
  type: push
  connect:
    type: tls
    address: "xxx.org:12345"
    ca: /etc/zrepl/ca
    cert: /etc/zrepl/cert
    key:  /etc/zrepl/key
    server_cn: "xxx"
  filesystems: {
    "pool/encrypted/test2": true
  }
  send:
    encrypted: true
    send_properties: true
  snapshotting:
    type: cron
    prefix: zrepl_cron_odd_
    cron: "1/2 13,14,15 * * *"
  pruning:
    keep_sender:
    - type: regex
      regex: ".*"
    keep_receiver:
    - type: grid
      grid: 1x1h(keep=all) | 24x1h | 30x1d
      regex: "^zrepl_.*"
    - type: regex
      negate: true
      regex: "^zrepl_.*"

- name: TLS2
  type: push
  connect:
    type: tls
    address: "xxx.org:12345"
    ca: /etc/zrepl/ca
    cert: /etc/zrepl/cert
    key:  /etc/zrepl/key
    server_cn: "xxx"
  filesystems: {
    "pool/encrypted/test2": true
  }
  send:
    encrypted: true
    send_properties: true
  snapshotting:
    type: cron
    prefix: zrepl_cron_even_
    cron: "*/2 13,14,15 * * *"
  pruning:
    keep_sender:
    - type: regex
      regex: ".*"
    keep_receiver:
    - type: grid
      grid: 1x1h(keep=all) | 24x1h | 30x1d
      regex: "^zrepl_.*"
    - type: regex
      negate: true
      regex: "^zrepl_.*"

```

problame:
Thanks for testing this again. Will go ahead and merge.

PR #615, opened 2022-07-11
Use $(MAKE) instead of make

gmekicaxcient:
Building on FreeBSD on should execute `gmake` not `make`, fixing it by using $(MAKE).


problame:
Thanks for your contribution!

PR #619, opened 2022-07-18
Filter bookmarks when searching filesystem versions.

### LINKED ISSUE #490: Chained replication jobs conflicting
I have a pretty unusual setup with three servers doing "chained replication" of zfs filesystems.

By "chained replication" I mean the following scenario:
Consider servers A,B and C. The zfs filesystem from A is replicated using zrepl to B using a snap job on A and a pull job on B. The zfs filesystem on B is replicated using zrepl to C using a snap job on B and a pull job on C.

The conflict arises because the pull job on B finds the bookmark created on the filesystem by the pull job running on C. This bookmark obviously cannot be found on A so the job considers the two filesystems diverged and the conflict cannot be resolved.

`Jun  7 12:02:36 backup01 zrepl[9348]: [nxtmail01_sync][repl][Uv38$Ny6V$7Bc5$7Bc5.6hqC]: conflict conflict="the receiver's latest snapshot is not present on sender:\n  last common: @zrepl_nxtm01-b01_20210606_140303_000\n  sender-only:\n  receiver-only:\n    @zrepl_b01-b03_20210607_090523_000\n" filesystem="mail/vmail"`

I have drilled this problem down to the endpoint.go implementation and have come up with a very dirty fix by filtering out the listing of bookmarks in the Receiver's ListFileSystemVersions method. I realize this is not a usable fix for any other use case, it could in fact cause problems, this is the reason I didn't create a PR. I'm attaching a patch.

[zrepl_conflict_dirty_fix.patch.txt](https://github.com/zrepl/zrepl/files/6608954/zrepl_conflict_dirty_fix.patch.txt)

The way I see it, the bookmark causing the problems does contain the name of the job that created it. Maybe a good solution would be to somehow process only bookmarks created by the job currently running and disregarding the rest.

I do not understand the inner working of zrepl enough to be able to go further fixing this, but if I could get some pointers on how to proceed further I'm willing to help fix the issue.

kresike:
Fixes #490

This is an old bug that I found and have been using this fix since. I've updated it to the newest version and seems to be working fine. @kpfleming please test it if you have the resources for it.

@problame please review.

### DIFF
--- a/replication/logic/diff/diff.go
+++ b/replication/logic/diff/diff.go
@@ -88,7 +88,26 @@ func SortVersionListByCreateTXGThenBookmarkLTSnapshot(fsvslice []*FilesystemVers
 	return sorted
 }
 
+func StripBookmarksFromVersionList(fsvslice []*FilesystemVersion) []*FilesystemVersion {
+	fslice := make([]*FilesystemVersion, 0, len(fsvslice))
+	for _, fv := range fsvslice {
+		if fv.Type != FilesystemVersion_Bookmark {
+			fslice = append(fslice, fv)
+		}
+	}
+	return fslice
+}
+
 func IncrementalPath(receiver, sender []*FilesystemVersion) (incPath []*FilesystemVersion, conflict error) {
+
+	// Receive-side bookmarks can't be used as incremental-from,
+	// and don't cause recv to fail if there is a newer bookmark than incremetal-form on the receiver.
+	// So, simply mask them out.
+	// This will also hide them in the report, but it keeps the code in this function simple,
+	// and a user who complains about them missing in a conflict message will likely require
+	// more education about bookmarks than a slightly more accurate error message. They'll get
+	// that when they open an issue.
+	receiver = StripBookmarksFromVersionList(receiver)
 
 	receiver = SortVersionListByCreateTXGThenBookmarkLTSnapshot(receiver)
 	sender = SortVersionListByCreateTXGThenBookmarkLTSnapshot(sender)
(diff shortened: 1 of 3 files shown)


kpfleming:
Thanks @kresike! I'll test this in the next few days.

replication/logic/diff/diff.go
```diff
@@ -88,8 +88,20 @@ func SortVersionListByCreateTXGThenBookmarkLTSnapshot(fsvslice []*FilesystemVers
...
+		if fv.Type != FilesystemVersion_Bookmark {
+			fslice = append(fslice, fv)
+		}
+	}
+	return fslice
+}
+
 func IncrementalPath(receiver, sender []*FilesystemVersion) (incPath []*FilesystemVersion, conflict error) {
 
+	receiver = FilterBookmarksFromVersionList(receiver)
```
problame, line 103: 1. Naming: let's call this `StripBookmarksFromVersionList` 
2. Please add a comment to explain why we need to strip the bookmarks.

replication/logic/diff/diff.go
```diff
@@ -88,8 +88,20 @@ func SortVersionListByCreateTXGThenBookmarkLTSnapshot(fsvslice []*FilesystemVers
 	return sorted
 }
 
+func FilterBookmarksFromVersionList(fsvslice []*FilesystemVersion) []*FilesystemVersion {
+	fslice := make([]*FilesystemVersion, 0)
```
problame, line 92: nit: you can use `make(..., 0, len(fsvslice))` to suggest a capacity while still keeping length at 0. Makes for fewer allocations.

problame (changes requested):
Thanks for your contribution, @kresike !

problame:
I'm going to merge this, as it definitely fixes a problem, and I assume @kresike tested it on their side.

Added a comment, unit tests, and changelog entry.
Thanks for your contribution, @kresike

PR #653, opened 2022-12-05
Report if dataset does not exist but is in configuration

gmekicaxcient:
The idea is to report missing datasets through Prometheus. I used different metrics for different job types as it is easier to measure and debug in development, so it's an open question if it would be better to use one metric for all job types.

As this is my first open source project written in Go, please do tell me how to improve this code if you have any suggestions.

### DIFF
--- a/dist/grafana/grafana-prometheus-zrepl.json
+++ b/dist/grafana/grafana-prometheus-zrepl.json
@@ -1152,7 +1152,7 @@
                 },
                 {
                   "color": "red",
-                  "value": 80
+                  "value": 0
                 }
               ]
             }
@@ -1185,7 +1185,7 @@
               "uid": "${DS_PROMETHEUS}"
             },
             "editorMode": "builder",
-            "expr": "zrepl_zfs_list_unmatched_user_specified_dataset_count",
+            "expr": "rate(zrepl_zfs_list_unmatched_user_specified_dataset_count[1m])",
             "legendFormat": "__auto",
             "range": true,
             "refId": "A"
--- a/dist/grafana/grafana-prometheus-zrepl.json
+++ b/dist/grafana/grafana-prometheus-zrepl.json
@@ -1,1384 +1,1513 @@
 {
-    "__inputs": [
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__elements": {},
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "9.3.6"
+    },
+    {
+      "type": "panel",
+      "id": "graph",
+      "name": "Graph (old)",
+      "version": ""
+    },
+    {
+      "type": "panel",
+      "id": "heatmap",
+      "name": "Heatmap",
+      "version": ""
+    },
+    {
+      "type": "datasource",
+      "id": "prometheus",
+      "name": "Prometheus",
+      "version": "1.0.0"
+    },
+    {
+      "type": "panel",
+      "id": "stat",
+      "name": "Stat",
+      "version": ""
+    },
+    {
+      "type": "panel",
+      "id": "table",
+      "name": "Table",
+      "version": ""
+    },
+    {
+      "type": "panel",
+      "id": "text",
+      "name": "Text",
+      "version": ""
+    }
+  ],
+  "annotations": {
+    "list": [
       {
-        "name": "DS_PROMETHEUS",
-        "label": "Prometheus",
-        "description": "",
-        "type": "datasource",
-        "pluginId": "prometheus",
-        "pluginName": "Prometheus"
-      }
-    ],
-    "__elements": {},
-    "__requires": [
-      {
-        "type": "grafana",
-        "id": "grafana",
-        "name": "Grafana",
-        "version": "9.3.6"
-      },
-      {
-        "type": "panel",
-        "id": "graph",
-        "name": "Graph (old)",
-        "version": ""
-      },
-      {
-        "type": "datasource",
-        "id": "prometheus",
-        "name": "Prometheus",
-        "version": "1.0.0"
-      },
-      {
-        "type": "panel",
-        "id": "stat",
-        "name": "Stat",
-        "version": ""
-      },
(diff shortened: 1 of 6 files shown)


problame (changes requested):
My understanding of this PR is that you want a metric for the number of datasets that the configuration mentions by name, but are not present.
I can see a need for that and would merge such a feature.

As per your proposed changes, here is some general feedback:

1. We already evaluate the filesystem filter during replication:
here
https://github.com/zrepl/zrepl/blob/a6aa610165c60328f951cfc05c6429a531431bda/endpoint/endpoint.go#L615
and here
https://github.com/zrepl/zrepl/blob/a6aa610165c60328f951cfc05c6429a531431bda/endpoint/endpoint.go#L102

I would rather we extend the `DatasetFilter` interface so that, in addition to filter evaluation, it allows reporting the number of rules that haven't matched at all. You can then turn that into a metric.

2. You added a separate metric per job type (push, pull, ...). How about you just turn the job type into a prometheus label instead? 

So, with 1 and 2 combined, the metric would look like `zrepl_unmatched_dataset_filter_rule_count{jobType="push|pull|...", jobName="myjobname"} VALUE` . Please search for precedence in other metrics, I'm not sure whether `jobType` is commonly used vs something else.

gmekicaxcient:
Sorry for late reply! I do understand now what needs to be done, give me some time to experiment. Thank you!

gmekicaxcient:
I think I covered what you explained, but I'm still new to the codebase so please do give hints if something else needs working on. The metric has only "zrepl_job" label as I think job type wouldn't help much, so do tell me if other labels would be good to have here. Also, zrepl_job is the label used in other metrics so I used it, too.

gmekicaxcient:
I changed `Add` to `Set` as number of errors just keeps growing. I hope that is fine. Otherwise I didn't find any errors in scenarios I tried out.

gmekicaxcient:
I was thinking, maybe it would be good to have prometheus metrics for missing datasets reported through logger, too. The reasoning behind it is that not everybody have prometheus configured, but they do want to know if there's a missconfiguration. What do you think?

zfs/mapping.go
```diff
@@ -3,12 +3,18 @@ package zfs
 import (
 	"context"
 	"fmt"
+
+	"github.com/zrepl/zrepl/zfs/zfscmd"
 )
 
 type DatasetFilter interface {
 	Filter(p *DatasetPath) (pass bool, err error)
+	UserSpecifiedDatasets() UserSpecifiedDatasetsSet
```
problame, line 12:
```suggestion
	// The caller owns the returned set.
	// Implementations should return a copy.
	UserSpecifiedDatasets() UserSpecifiedDatasetsSet
```

zfs/mapping.go
```diff
@@ -87,5 +97,9 @@ func ZFSListMappingProperties(ctx context.Context, filter DatasetFilter, propert
 
 	}
 
+	jobid := zfscmd.GetJobIDOrDefault(ctx, "__nojobid")
+	metric := prom.ZFSListUnmatchedUserSpecifiedRulesCount.WithLabelValues(jobid)
+	metric.Set(float64(len(unmatchedUserSpecifiedDatasets)))
```
problame, line 102: This needs to be a counter that is incremented using `.Add()`.

The reason is that, at the zfs package level, we don't that the endpoint.go:ListFilesystems function is the only caller.

So, if we just `Set()` it to whatever the latest result was, it could flip if there are two callers, one with tan unmatched dataset, one without.

The solution is to use `Add()` with a counter, and use Prometheus to alert on "did the counter increase in the last $timeframe".

zfs/prometheus.go
```diff
@@ -34,6 +35,14 @@ func init() {
 		Name:      "destroy_duration",
 		Help:      "Duration it took to destroy a dataset",
 	}, []string{"dataset_type", "filesystem"})
+	prom.ZFSListUnmatchedUserSpecifiedRulesCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "zrepl",
+		Subsystem: "zfs",
+		Name:      "list_unmatched_user_specified_rules_count",
```
problame, line 41: Lets rename the metric and the variable name to

```
zrepl_zfs_list_unmatched_user_specified_datasets_count
```

That gets it in line with the name of the interface method.

problame (changes requested):
Responding to your question through this review.

problame:
Pushed a commit that cleaned up a few remaining things.

Now the remaining question is: do you want to augment the Grafana Dashboard with a panel for this new metrics?
It would demonstrate the usefulness of this feature, and I figure you already have something like this on your internal systems anyways.

gmekicaxcient:
Which dashboard is that? I mean, we still don't have this metric internally as patching existing version with this PR is not easy, so I have to ask.

gmekicaxcient:
Ah, you mean [this Grafana Dashboard](https://github.com/zrepl/zrepl/blob/master/dist/grafana/grafana-prometheus-zrepl.json). Yes, I'll make that, too.

problame:
I squashed & rebased this PR onto the latest `master` where I updated the Grafana dashboard to Grafana 9.3.6 in the meantime. That makes the diff smaller.

Will post a review.

dist/grafana/grafana-prometheus-zrepl.json
```diff
@@ -1093,6 +1102,98 @@
...
+            "thresholds": {
+              "mode": "absolute",
+              "steps": [
+                {
+                  "color": "green",
+                  "value": null
+                },
+                {
+                  "color": "red",
+                  "value": 80
```
problame, line 1155: The metric is a counter that only goes up, except on zrepl restart. So, we need to alert on some derivative of the value, so that we go back to green when everything is fine again.

problame (changes requested):

dist/grafana/grafana-prometheus-zrepl.json
gmekicaxcient: I played with different functions but in the end I think `rate()` works the best as you can see spikes in the graph where raise means one or more new unmatched datasets, fall means one or more datasets are matched (either removed from config or datasets are created).

gmekicaxcient:
Anything else I can improve?

problame:
Posted an improved dashboard:

image

I think this looks good, will merge it

problame (approved):

problame:
Thanks for the patience, @gmekicaxcient  !

PR #664, opened 2022-12-24
Include openrc service file

gramosg:
This PR includes a simple service file for openrc-based distributions such as Gentoo or Alpine Linux.

### DIFF
--- a/dist/openrc/zrepl
+++ b/dist/openrc/zrepl
@@ -7,8 +7,17 @@ pidfile="/run/${RC_SVCNAME}.pid"
 output_log="/var/log/${RC_SVCNAME}.log"
 error_log="/var/log/${RC_SVCNAME}.log"
 
+zrepl_runtime_dir='/var/run/zrepl'
+
 start() {
-	mkdir -p /var/run/zrepl
-	chmod 750 /var/run/zrepl
+	mkdir -p $zrepl_runtime_dir/stdinserver
+	chmod -R 0700 $zrepl_runtime_dir
 	default_start
 }
+
+stop() {
+	rm -rf $zrepl_runtime_dir
+	default_stop
+}
+
+# vi: noet sw=8 sts=0
--- a/dist/openrc/zrepl
+++ b/dist/openrc/zrepl
@@ -1,0 +1,14 @@
+#!/sbin/openrc-run
+
+command='/usr/local/bin/zrepl'
+command_args='daemon'
+command_background='true'
+pidfile="/run/${RC_SVCNAME}.pid"
+output_log="/var/log/${RC_SVCNAME}.log"
+error_log="/var/log/${RC_SVCNAME}.log"
+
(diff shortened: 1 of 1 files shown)


dist/openrc/zrepl
```diff
@@ -0,0 +1,14 @@
+#!/sbin/openrc-run
+
+command='/usr/local/bin/zrepl'
```
problame, line 3: Why not `/usr/bin/zrepl` ? That's where the .deb and .rpm packages install it.

gramosg: @problame I've just copied the path from the [systemd service](https://github.com/zrepl/zrepl/blame/master/dist/systemd/zrepl.service#L8) already present in this repo.

Also, if someone is copying the service from this repository instead of using the one provided by the distro it probably means zrepl was installed manually so it should be located in /usr/local/bin.

problame: Allright.

dist/openrc/zrepl
```diff
@@ -0,0 +1,14 @@
...
+command_args='daemon'
+command_background='true'
+pidfile="/run/${RC_SVCNAME}.pid"
+output_log="/var/log/${RC_SVCNAME}.log"
+error_log="/var/log/${RC_SVCNAME}.log"
+
+start() {
+	mkdir -p /var/run/zrepl
+	chmod 750 /var/run/zrepl
```
problame, line 12: For the systemd file, we use

```
RuntimeDirectory=zrepl zrepl/stdinserver
RuntimeDirectoryMode=0700
```

It creates both `/var/run/zrepl` and `/var/run/zrepl/stdinserver`.
The latter is relevant for the `ssh` transport.
I think the openrc file should create it as well.

Also, systemd auto-removes the runtime directory when stopping the unit.
And it recreates it when restarting the unit.
I don't think zrepl depends on this behavior though - it will re-bind unix socket files if they already exist.

problame (changes requested):

dist/openrc/zrepl
gramosg: @problame Thanks for the feedback, I've fixed both issues (missing stdinserver + runtime directory removal when stopping). Also changed permissions from 750 to 700 to keep it the same as the systemd unit, although I think both are sensible options.

problame (changes requested):
please quote `$zrepl_runtime_dir`, in particular the one after `rm -rf` ;)

problame (approved):

problame:
Thanks for sticking with this!

I'll merge this PR now, but, what are you planning to do once this is merged?

gramosg:
Thanks to you for the thorough review! I'm not planning to do anything special after the merge, I just wrote a service that worked for me and wanted to contribute it in case it's useful for someone else. My home NAS runs Gentoo and "zrepls" to the backup machine :)

Issue #165, opened 2019-03-28
zrepl status: the progress bar never finish but the job is complete


problame:
* Another detail in the replication job, the progress bar never finish but the job is completed:

```
Job: zfsnode                                                                                                                                                                                                                                   
    Type: push                                                                                                                                                                                                                                  
    Replication:                                                                                                                                                                                                                                
        Status: Completed                                                                                                                                                                                                                       
        Progress: [=================================================>-] 59.6 GiB / 60.4 GiB @ 0 B/s                                                                                                                                             
          APool/px Completed (step 1/1, 29.7 GiB/30.1 GiB)                                                                                                                                                                               
          BPool/px Completed (step 1/1, 30.0 GiB/30.3 GiB)  
```

-- originally reported as part of https://github.com/zrepl/zrepl/issues/162#issue-425513076

problame:
This situation is the result of under-estimation by `zfs send -n`, which can happen. So, `Status: Completed` should just force progress to 100% iff no errors occured during replication.

lapo-luchini:
Me, on the other hand, seem to always have the opposite effect?
When job is ended my progress is always more than expected.

```
Job: backup
    Type: push
    Replication:
        Attempt #1
        Status: done
        Progress: [==================================================>] 526.3 MiB / 487.6 MiB @ 0 B/
```

problame:
@lapo-luchini upcoming release 0.3 notifies about _under_-estimation (ZFS currently cannot make size estimates for from=bookmark sends). See 9a4763ceee77b4a8541a518ab23bdb982b7f137e

Any testing of the current `master` branch is greatly appreciated!

problame:
Late follow up regarding under-estimation for bookmarks: https://github.com/zrepl/zrepl/issues/649 

I think the solution for this issue is to hide the progress bar once all filesystems are in a terminal state.
And if the size estimate was off by more than X, report it in a line.
Will rename the issue.

PR #674, opened 2023-01-24
Fix issue #165. zrepl status: hide progress bar once all filesystems reach terminal state

0x3333:
Hide the progress bar once all filesystems reach terminal state on pruner.

Before:
image

After:
image

### DIFF
--- a/client/status/viewmodel/render.go
+++ b/client/status/viewmodel/render.go
@@ -374,13 +374,15 @@ func renderReplicationReport(t *stringbuilder.B, rep *report.Report, history *by
 			eta = time.Duration((float64(expected)-float64(replicated))/float64(rate)) * time.Second
 		}
 
-		t.Write("Progress: ")
-		t.DrawBar(50, replicated, expected, changeCount)
-		t.Write(fmt.Sprintf(" %s / %s @ %s/s", ByteCountBinaryUint(replicated), ByteCountBinaryUint(expected), ByteCountBinary(rate)))
-		if eta != 0 {
-			t.Write(fmt.Sprintf(" (%s remaining)", humanizeDuration(eta)))
+		if latest.State != report.AttemptDone {
+			t.Write("Progress: ")
+			t.DrawBar(50, replicated, expected, changeCount)
+			t.Write(fmt.Sprintf(" %s / %s @ %s/s", ByteCountBinaryUint(replicated), ByteCountBinaryUint(expected), ByteCountBinary(rate)))
+			if eta != 0 {
+				t.Write(fmt.Sprintf(" (%s remaining)", humanizeDuration(eta)))
+			}
+			t.Newline()
 		}
-		t.Newline()
 		if containsInvalidSizeEstimates {
 			t.Write("NOTE: not all steps could be size-estimated, total estimate is likely imprecise!")
 			t.Newline()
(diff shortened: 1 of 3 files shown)


problame:
Thanks for fixing this

I don't get why the CI is failing.
1. Did it prompt you for anything specific? 
2. Please rebase & force-push to trigger it again.

0x3333:
> I don't get why the CI is failing.

I was not logged in CircleCI 🤦🏻‍♂️... rebased and forced-push. Let's see.

client/status/viewmodel/render.go
```diff
@@ -493,15 +495,17 @@ func renderPrunerReport(t *stringbuilder.B, r *pruner.Report, fsfilter FilterFun
...
-	t.Write(">")
-	t.Write(stringbuilder.Times("-", 80-progress))
-	t.Write("]")
-	t.Printf(" %d/%d snapshots", completedDestroyCount, totalDestroyCount)
-	t.Newline()
+	if state != pruner.Done {
+		progress := int(math.Round(80 * float64(completedDestroyCount) / float64(totalDestroyCount)))
+		t.Write("Progress: ")
+		t.Write("[")
+		t.Write(stringbuilder.Times("=", progress))
```
InsanePrawn, line 502: @problame any reason why this doesn't use `t.DrawBar()` like the code above?

client/status/viewmodel/render.go
```diff
@@ -374,13 +374,15 @@ func renderReplicationReport(t *stringbuilder.B, rep *report.Report, history *by
 			eta = time.Duration((float64(expected)-float64(replicated))/float64(rate)) * time.Second
 		}
 
-		t.Write("Progress: ")
-		t.DrawBar(50, replicated, expected, changeCount)
-		t.Write(fmt.Sprintf(" %s / %s @ %s/s", ByteCountBinaryUint(replicated), ByteCountBinaryUint(expected), ByteCountBinary(rate)))
-		if eta != 0 {
-			t.Write(fmt.Sprintf(" (%s remaining)", humanizeDuration(eta)))
+		if latest.State != report.AttemptDone {
```
problame, line 377: States `AttemptFanOutError` and `AttemptPlanningError` are also terminal states where displaying a progress bar doesn't make sense.

Consider adding an `IsTerminal()` method to `AttemptState`

Here's the code that sets the state:

https://github.com/zrepl/zrepl/blob/b9250a41a25796969246710e55c972f24d1284d6/replication/driver/replication_driver.go#L698-L715

client/status/viewmodel/render.go
```diff
@@ -493,15 +495,17 @@ func renderPrunerReport(t *stringbuilder.B, r *pruner.Report, fsfilter FilterFun
...
-	progress := int(math.Round(80 * float64(completedDestroyCount) / float64(totalDestroyCount)))
-	t.Write("Progress: ")
-	t.Write("[")
-	t.Write(stringbuilder.Times("=", progress))
-	t.Write(">")
-	t.Write(stringbuilder.Times("-", 80-progress))
-	t.Write("]")
-	t.Printf(" %d/%d snapshots", completedDestroyCount, totalDestroyCount)
-	t.Newline()
+	if state != pruner.Done {
```
problame, line 498: Similarly, for pruner, terminal states are  `PlanErr`, `ExecErr` and `Done`

Learn more by checking the places where `doOneAttempt` sets the state and returns

https://github.com/zrepl/zrepl/blob/1f0f2f856966713c06b76943d9de0c560f820557/daemon/pruner/pruner.go#L352

problame (changes requested):
Thanks for working on this. I think we should also hide the status bar in terminal error states, not just in terminal success states. WDYT?

client/status/viewmodel/render.go
problame: Just an oversight, I think. @0x3333, feel free to fix this in this PR.

0x3333: I just added the `if`(😅), but I updated it to use DrawBar. I used `changeCount` equal `1` so the progress doesn't show as stalled.

client/status/viewmodel/render.go
0x3333: I added the `IsTerminal` method. Nice catch

client/status/viewmodel/render.go
0x3333: Same as `AttemptState`.

PR #687, opened 2023-03-11
keep_bookmarks option was removed

skirmess:
I found the problem regarding ``keep_bookmarks`` confusing and scary but couldn't find further information about this option in the documentation because the option was removed since the talks were given.

This small change adds a note that the keep_bookmarks option was removed in 0.4.0.


problame:
Thanks for this patch. I made some corrections on top, will merge it now.

Issue #742, opened 2023-10-08
Encrypted send to untrusted target not quite working ... pretty sure it's a simple config issue


Halfwalker:
Encrypted send to untrusted target not quite working ...

I have a backup dataset `bondi3/zrepl` on a server **bondi** that receives laptop backups.  The dataset is encrypted, and laptop sends are plain, so they're re-encrypted on **bondi**.  For offsite backups, I want to replicate this to a backup dataset `wider/zrepl` on remote server **widey**.  Should be fairly simple ...

The two laptop sub-datasets under `bondi3/zrepl` - both look like this
```
bondi3/zrepl/ryzen2/ryzen22/home/alice  encryption             aes-256-gcm            -
bondi3/zrepl/ryzen2/ryzen22/home/alice  keylocation            none                   default
bondi3/zrepl/ryzen2/ryzen22/home/alice  keyformat              passphrase             -
bondi3/zrepl/ryzen2/ryzen22/home/alice  pbkdf2iters            350000                 -
bondi3/zrepl/ryzen2/ryzen22/home/alice  encryptionroot         bondi3/zrepl 
bondi3/zrepl/ryzen2/ryzen22/home/alice  keystatus              available
```

**bondi** zrepl.yml *source* job
```
  - type: source
    name: laptop_source
    filesystems:
      "bondi3/zrepl<": true
    snapshotting:
      type: manual
    send:
      encrypted: true
    serve:
      type: tls
      listen: :8541
      ca: /etc/zrepl/ca.crt
      cert: /etc/zrepl/bondi.crt
      key: /etc/zrepl/bondi.key
      client_cns:
        - widey
```

**widey** zrepl.yml *pull* job
```
  - name: pull_bondi
    type: pull
    connect:
      type: tls
      address: 192.168.2.104:8541
      ca: /etc/zrepl/ca.crt
      cert: /etc/zrepl/widey.crt
      key: /etc/zrepl/widey.key
      server_cn: bondi
    root_fs: wider/zrepl
    interval: 1h
    replication:
      protection:
        initial: guarantee_resumability
        incremental: guarantee_resumability
    recv:
      properties:
        override: {
          "readonly": "on",
          "mountpoint": "none",
          "canmount": "noauto"
        }
      # encryption off as per encrypted-send-to-untrusted-receiver
      # https://zrepl.github.io/master/configuration/sendrecvoptions.html#placeholders
      placeholder:
        encryption: off
    pruning:
      keep_sender:
        - type: regex
[...]

PR #744, opened 2023-10-14
fix: replication of placeholder filesystems

problame:
fixes https://github.com/zrepl/zrepl/issues/742

Before this PR, when chaining replication from
A => B => C, if B had placeholders and the `filesystems`
included these placeholders, we'd incorrectly
fail the planning phase with error
`sender does not have any versions`.

The non-placeholder child filesystems of these placeholders
would then fail to replicate because of the
initial-replication-dependency-tracking that we do, i.e.,
their parent failed to initially replication, hence
they fail to replicate as well
(`parent(s) failed during initial replication`).

We can do better than that because we have the information
whether a sender-side filesystem is a placeholder.
This PR makes the planner act on that information.
The outcome is that placeholders are replicated as
placeholders (albeit the receiver remains in control
of how these placeholders are created, i.e., `recv.placeholders`)
The mechanism to do it is:
1. Don't plan any replication steps for filesystems that
   are placeholders on the sender.
2. Ensure that, if a receiving-side filesystem exists, it
   is indeed a placeholder.

Check (2) may seem overly restrictive, but, the goal here
is not just to mirror all non-placeholder filesystems, but
also to mirror the hierarchy.

Testing performed:
- [x] confirm with issue reporter that this PR fixes their issue
- [x] add a regression test that fails without the changes in this PR


problame:
Tested & found working in https://github.com/zrepl/zrepl/issues/742

What's still needed is a regression test

Issue #691, opened 2023-03-18
`type: last_n` ignores `regex` and keeps every snapshot if less than `count` snapshots exist


skirmess:
The `last_n` prune rule keeps everything, regardless of if it matches the `regex` or not, if there are less than `count` snapshot. The expectation would be to never keep non-`regex` snapshots, regardless of number.

#### I have the following config:

```
jobs:
- name: snapshots
  type: snap
  filesystems: {
    'rpool/enc/zrepl-test3': true
  }
  snapshotting:
    type: cron
    cron: '* * * * *'
    prefix: snapshot_
    timestamp_format: 2006-01-02-150405Z
  pruning:
    keep:
      - type: last_n
        count: 6
        regex: '\Asnapshot_'
```

#### After the first run it looks like this:

zrepl creates a snapshot.
```
$ zfs list -H -t snapshot -d 1 rpool/enc/zrepl-test3
rpool/enc/zrepl-test3@snapshot_2023-03-19-115800Z       0       -       336K    -
```

#### Create some snapshots:

We create manually some snapshots. There is nothing that keeps these snapshots, they should be removed the next run.

```
$ pfexec zfs snapshot rpool/enc/zrepl-test3@test1
$ pfexec zfs snapshot rpool/enc/zrepl-test3@test2
```

#### After the next run it looks like this:

Because there are less than 6 (`count`) snapshots, zrepl keeps all of them - ignoring the `regex`.

```
$ zfs list -H -t snapshot -d 1 rpool/enc/zrepl-test3
rpool/enc/zrepl-test3@snapshot_2023-03-19-115800Z       0       -       336K    -
rpool/enc/zrepl-test3@test1     0       -       336K    -
rpool/enc/zrepl-test3@test2     0       -       336K    -
rpool/enc/zrepl-test3@snapshot_2023-03-19-115900Z       0       -       336K    -
```

#### Then like this:

The next run creates a new snapshot, increasing the number of snapshots to 5. All of them are kept.

```
 zfs list -H -t snapshot -d 1 rpool/enc/zrepl-test3
rpool/enc/zrepl-test3@snapshot_2023-03-19-115800Z       0       -       336K    -
rpool/enc/zrepl-test3@test1     0       -       336K    -
rpool/enc/zrepl-test3@test2     0       -       336K    -
rpool/enc/zrepl-test3@snapshot_2023-03-19-115900Z       0       -       336K    -
[...]

PR #750, opened 2023-10-18
Fix last_n keep rule (#691)

dsh2dsh:
From https://github.com/zrepl/zrepl/issues/691

The last_n prune rule keeps everything, regardless of if it matches the regex or not, if there are less than count snapshot. The expectation would be to never keep non-regex snapshots, regardless of number.

### DIFF
--- a/pruning/keep_last_n_test.go
+++ b/pruning/keep_last_n_test.go
@@ -90,7 +90,7 @@ func TestKeepLastN(t *testing.T) {
 				stubSnap{"a2", false, o(12)},
 			},
 			rules: []KeepRule{
-				MustKeepLastN(3, "a"),
+				MustKeepLastN(4, "a"),
 			},
 			expDestroy: map[string]bool{
 				"b1": true,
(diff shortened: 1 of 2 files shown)


pruning/keep_last_n_test.go
```diff
@@ -90,7 +90,7 @@ func TestKeepLastN(t *testing.T) {
 				stubSnap{"a2", false, o(12)},
 			},
 			rules: []KeepRule{
-				MustKeepLastN(3, "a"),
+				MustKeepLastN(4, "a"),
```
problame, line 93: Why this change?

problame (changes requested):
Thanks for this contribution, and sorry for the late reply!

This PR is correct, and in line what the docs always said.

Usually with this kind of change there's risk about existing setups relying on the bug, but, with this one, I can't see a problem. So, let's just fix it.

I'd love for more tet coverage. Can you add ad additional test case that fails without this PR and works with this PR?

dsh2dsh:
> Can you add ad additional test case that fails without this PR and works with this PR?

Already done. This line
```go
MustKeepLastN(4, "a")
```
tests it.

pruning/keep_last_n_test.go
dsh2dsh: It tests `keep_last_n` works correctly in situation, where number of snapshots `< n`. See https://github.com/zrepl/zrepl/issues/691 for details.

problame:
Thanks again, and sorry for the long delay!

PR #752, opened 2023-10-19
Allow same root_fs for different jobs: sinks and so on

dsh2dsh:
Because some jobs add client identity to root_fs and other jobs don't do that,
we can't reliable detect overlapping of filesystems. And and the same time we
need an ability to use equal or overlapped root_fs for different jobs. For
instance see this config:

```
  - name: "zdisk"
    type: "sink"
    root_fs: "zdisk/zrepl"
    serve:
      type: "local"
      listener_name: "zdisk"
```
and
```
  - name: "remote-to-zdisk"
    type: "pull"
    connect:
      type: "tls"
    root_fs: "zdisk/zrepl/remote"
```

As you can see, two jobs have overlapped root_fs, but actually datasets are not
overlapped, because job `zdisk` save everything under `zdisk/zrepl/localhost`,
because it adds client identity. So they actually use two different filesystems:
`zdisk/zrepl/localhost` and `zdisk/zrepl/remote`. And we can't detect this
situation during config check. So let's just remove this check, because it's
admin's duty to configure correct root_fs's.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,19 +58,26 @@ commands:
             git config --global user.email "zreplbot@cschwarz.com"
             git config --global user.name "zrepl-github-io-ci"
 
-      # https://circleci.com/docs/2.0/add-ssh-key/#adding-multiple-keys-with-blank-hostnames
-      - run: ssh-add -D
-      # the default circleci ssh config only additional ssh keys for Host !github.com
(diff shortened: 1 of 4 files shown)


problame:
Will merge this after https://github.com/zrepl/zrepl/pull/763

Issue #601, opened 2022-05-07
"all pools" filter


clarfonthey:
Right now, it's possible to request backing up all pools by providing the `>` pattern to zrepl, but it's not possible to disable patterns based upon datasets within pools.

Essentially, I recently reorganised my pools such that the entire pools are snapshotted, except for the datasets within a `nosnap` dataset at the root of the pools. It would be helpful to be able to annotate this via zrepl, using (syntax bikesheddable):

```
filesystems:
    "/<": true
    "/nosnap<": false
```

where here, `/` indicates "all pools".

Alternatively, a `*` pattern could be added that represents any pool or dataset. For example, you could do `*/nosnap` to mean "the `nosnap` dataset in any pool" or `pool/*/nosnap` to mean "the `nosnap` dataset within any dataset inside the pool `pool`". This would also extend to `**`, which could match multiple levels of datasets.

Of course, I understand the desire to avoid adding such wildcards, since they could make the lookup logic for zrepl much more complicated; this is why I propose just adding an "all pools" wildcard for now, since it at least helps solve the use case of using the same filesystem conventions on multiple pools, and/or wanting to copy the same config to multiple machines who may have differently named pools.

problame:
We already have a similar request at https://github.com/zrepl/zrepl/issues/287

The unanswered problem there, which I repeat, for you to answer for your feature request:

> I think the difficulty here lies in prioritization compared to the < 'subtree wildcard' and the non-wildcard filter rules.
(see https://zrepl.github.io/configuration/filter_syntax.html for the existing prioritization rules).
> In the current config syntax, the order of the filter rules does not matter and cannot be easily reconstructed due to constraints by the yaml parsing library.
> How would you go about this?

---

> , since they could make the lookup logic for zrepl much more complicated; 

What exactly are you referring to here?
I think figuring out which sender and dataset corresponds to which receiver dataset based on `receiver.root_fs + sender.name <==> sender.name`  could become tricky once we allow `**` style globs, but I don't think it would be a problem for `*` style globs.
What do you think? cc @InsanePrawn 

---

Note that this is much simpler than https://github.com/zrepl/zrepl/issues/253 and almost ready to implement, IMO.

clarfonthey:
So, I think that the most logical way is to go by depth-based specificity, then they're ordered lexicographically. So, from most to least specific, we'd have `path`, `*`, and `**`. When performing ordering, we can treat `path<` as an alias for `path/**`.

[...]

PR #755, opened 2023-10-21
Add support of shell patterns for filesystems

dsh2dsh:
Config like
```
filesystems:
  "zroot/jails</*/root": true
```
will include datasets `zroot/jails/a/root` and `zroot/jails/b/root`, but wont
include `zroot/jails/a` and `zroot/jails/b`.

### DIFF
--- a/daemon/filters/fsmapfilter.go
+++ b/daemon/filters/fsmapfilter.go
@@ -2,6 +2,7 @@ package filters
 
 import (
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


problame (commented):
The problem with wildcards is specificity, i.e., which rule to consider when multiple rules match. For example, take dataset `a/b/c/d` and rules `a/*/d` and `a/*/c/d`. Which rule is more specific? 

Also, did you think through what happens if there are multiple `*`  in one rule?


Also I think the

```
        "tank/bar</*/foo": true # rule 4
```

from the example looks quite weird.
Is there a reason why you only allow `*` for rules with `<`?

dsh2dsh:
> take dataset a/b/c/d and rules a/*/d and a/*/c/d. Which rule is more specific?

For one dataset one shell pattern rule possible. Somebody could add the same dataset twice with different shell patterns, but I think that's up to sysadmin, who is configuring zrepl.

> Also, did you think through what happens if there are multiple * in one rule?
It'll just works according to shell patterns: https://pkg.go.dev/path/filepath#Match

> from the example looks quite weird

Well, using of `<` in dataset looks same weird, but you use it. :)

> Is there a reason why you only allow * for rules with <?

Not sure I understand what do you mean. Shell patterns allow more symbols, see https://pkg.go.dev/path/filepath#Match

InsanePrawn (changes requested):
Whoops, had this review stuck unsent for weeks.

InsanePrawn:
I remember also writing a comment referencing #601, seems like that got lost though.

PR #756, opened 2023-10-21
Add file logging outlet

dsh2dsh:
Writes all log entries with minimum level `level` formatted by `format` to file from `filename`. This outlet automatically detects the log file was rotated and reopens it.

`time` and `log_level` are `true` by default.

Can be specified many times with different `filename`.

An example of configuration:
```
    - type: "file"
      level:  "info"
      format: "human"
      time: true
      log_level: true
      filename: "/var/log/zrepl.log"
```

Another example using `template`:
```
    - type: "file"
      level:  "info"
      format: "human"
      filename: "/var/log/zrepl.log"
      time: false
      template: '{{formatTime .Time "Jan _2 15:04:05"}} zrepl[{{.Pid}}]: {{.Message}}'
```

It formats log entries like
```
Oct 22 21:51:02 zrepl[29094]: [INFO][job-name][job][abcd$dcab$abcd.dcab]: wait for wakeups
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -421,6 +421,13 @@ type LoggingOutletCommon struct {
 	Format string `yaml:"format"`
 }
 
+type FileLoggingOutlet struct {
+	LoggingOutletCommon `yaml:",inline"`
+	FileName            string `yaml:"filename"`
+	Time                bool   `yaml:"time,default=true"`
+	LogLevel            bool   `yaml:"log_level,default=true"`
+}
+
 type StdoutLoggingOutlet struct {
 	LoggingOutletCommon `yaml:",inline"`
 	Time                bool `yaml:"time,default=true"`
@@ -582,6 +589,7 @@ func (t *SnapshottingEnum) UnmarshalYAML(u func(interface{}, bool) error) (err e
 
 func (t *LoggingOutletEnum) UnmarshalYAML(u func(interface{}, bool) error) (err error) {
 	t.Ret, err = enumUnmarshal(u, map[string]interface{}{
+		"file":   &FileLoggingOutlet{},
 		"stdout": &StdoutLoggingOutlet{},
 		"syslog": &SyslogLoggingOutlet{},
 		"tcp":    &TCPLoggingOutlet{},
--- a/config/config.go
+++ b/config/config.go
@@ -426,6 +426,7 @@ type FileLoggingOutlet struct {
 	FileName            string `yaml:"filename"`
 	Time                bool   `yaml:"time,default=true"`
 	LogLevel            bool   `yaml:"log_level,default=true"`
+	Template            string `yaml:"template"`
 }
 
 type StdoutLoggingOutlet struct {
--- a/daemon/logging/logging_outlets.go
+++ b/daemon/logging/logging_outlets.go
@@ -181,7 +181,6 @@ type FileOutlet struct {
 	filename  string
 	formatter EntryFormatter
 	template  *template.Template
-	writer    io.Writer
 }
 
 func (self *FileOutlet) WriteEntry(entry logger.Entry) error {
@@ -195,7 +194,10 @@ func (self *FileOutlet) WriteEntry(entry logger.Entry) error {
 	}
 
 	if self.template == nil {
-		return self.writeBytes(bytes)
+		if _, err := fmt.Fprintln(self.file, string(bytes)); err != nil {
+			return fmt.Errorf("failed write to %q: %w", self.filename, err)
+		}
+		return nil
 	}
 
 	if err := self.writeTemplate(entry.Time, string(bytes)); err != nil {
@@ -233,12 +235,11 @@ func (self *FileOutlet) reOpen() error {
 }
 
 func (self *FileOutlet) Open() error {
-	f, err := os.OpenFile(self.filename, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	f, err := os.OpenFile(self.filename, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
 	if err != nil {
 		return fmt.Errorf("file outlet: %w", err)
 	}
 	self.file = f
-	self.writer = f
 
 	return nil
 }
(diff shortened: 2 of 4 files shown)


daemon/logging/logging_outlets.go
```diff
@@ -170,5 +174,121 @@ func (o *SyslogOutlet) WriteEntry(entry logger.Entry) error {
...
+}
+
+func (self *FileOutlet) WriteEntry(entry logger.Entry) error {
+	bytes, err := self.formatter.Format(&entry)
+	if err != nil {
+		return err
+	}
+
+	if err := self.reOpenIfNotExists(); err != nil {
+		return nil
```
InsanePrawn, line 194: Shouldn't we be returning `err` here too?

docs/configuration/logging.rst
```diff
@@ -106,6 +106,54 @@ Outlets
...
+    * - ``filename``
+      - path of the log file
+    * - ``template``
+      - format output by Go template string
+
+Writes all log entries with minimum level ``level`` formatted by ``format`` to
+file from ``filename``. This outlet automatically detects the log file was
+rotated and reopens it.
+
+If ``template`` configured, log entries formatted by this template. For instance
```
InsanePrawn, line 139:
```suggestion
If ``template`` is configured, log entries are formatted using this template. For instance
```

InsanePrawn (changes requested):
Hi, thanks for the cool MR!

IMO, you've introduced two slightly separate things here:
- a new `file` log output target
- a new `template` format type

that we should probably split out from each other.

Is there any reason why the templating logic shouldn't be usable in e.g. the syslog outlet?

Also, `template:` should probably conflict with `format: json`, right? Making the template it's own `format` would do that.

daemon/logging/logging_outlets.go
dsh2dsh: Thanks, you are right. My typo.

dsh2dsh:
> that we should probably split out from each other.

My idea was to use the `template` for formatting log time mainly. So file outlet without `template` is useless for me. That's why I don't think it could be splitted.

Thinking about `template` as a formatter, I can't imagine how could it look like. Could you show an example of config for such formatter, how would you use it, what's use case?

> Is there any reason why the templating logic shouldn't be usable in e.g. the syslog outlet?

Yes. My idea was to use `template` for time formatting and adding PID. This has no reason for the syslog, because syslogd do that. I can't imagine how the `template` can be usable for syslog.

> Also, template: should probably conflict with format: json, right?

I wouldn't say conflict. May be not usable, but it's up to sysadmin. If somebody want it, why not. If somebody don't want use `template` and json formatter together, he just will not use it together.

> Making the template it's own format would do that.

I can't imagine it. What's its use case? How it would look like? In my understanding a formatter formats event attributes into a log line. I don't think this can be done using template language. Or may be I just don't see your idea. Could show an example?

daemon/logging/logging_outlets.go
```diff
@@ -170,5 +174,115 @@ func (o *SyslogOutlet) WriteEntry(entry logger.Entry) error {
...
+	return self.Open()
+}
+
+func (self *FileOutlet) Open() error {
+	f, err := os.OpenFile(self.filename, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return fmt.Errorf("file outlet: %w", err)
+	}
+	self.file = f
+	self.writer = f
```
problame, line 241: We're going to do unbuffered IO to the log file. That has a performance cost, especially for short log lines.

daemon/logging/logging_outlets.go
```diff
@@ -170,5 +174,115 @@ func (o *SyslogOutlet) WriteEntry(entry logger.Entry) error {
...
+
+	return nil
+}
+
+func (self *FileOutlet) ParseTemplate(templateText string) error {
```
problame, line 246: The `FileOutlet` is incomplete without parsing the templateText and setting self.template.

Also, the only caller is `parseFileOutlet`, which acts as the de-facto constructor function.

Please inline this function into `parseFileOutlet`

problame (changes requested):
Thanks for this contribution!

Left a few comments. Let's discuss 
https://github.com/zrepl/zrepl/pull/756#discussion_r1378205994 first.

config/config.go
```diff
@@ -421,6 +421,14 @@ type LoggingOutletCommon struct {
 	Format string `yaml:"format"`
 }
 
+type FileLoggingOutlet struct {
+	LoggingOutletCommon `yaml:",inline"`
+	FileName            string `yaml:"filename"`
+	Time                bool   `yaml:"time,default=true"`
+	LogLevel            bool   `yaml:"log_level,default=true"`
+	Template            string `yaml:"template"`
```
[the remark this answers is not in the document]
dsh2dsh: For configuration:
```
	flags := MetadataNone
	if in.Time {
		flags |= MetadataTime
	}
	if in.LogLevel {
		flags |= MetadataLevel
	}
```

daemon/logging/logging_outlets.go
```diff
@@ -170,5 +174,115 @@ func (o *SyslogOutlet) WriteEntry(entry logger.Entry) error {
...
+
+func (self *FileOutlet) WriteEntry(entry logger.Entry) error {
+	bytes, err := self.formatter.Format(&entry)
+	if err != nil {
+		return err
+	}
+
+	if err := self.reOpenIfNotExists(); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
dsh2dsh: > That's quite a lot of overhead.
Actually no. In real life there is no overhead at all. zrepl doesn't log hundred thousands messages per second.  And if, I repeat if, for somebody it'll have some overhead, somebody can just don't use it and use syslog or tcp instead.

> I guess because you're assuming there's external rotation?
Correct.

> we should include log rotation functionality
I don't think that's good idea. zrepl is for snapshots and newloslogd for log rotation. Every tool for single task.

daemon/logging/logging_outlets.go
dsh2dsh: I'm not sure I understand this comment. What do you mean?

daemon/logging/logging_outlets.go
dsh2dsh: I wouldn't like to add more lines into `parseFileOutlet` without real reason.

PR #758, opened 2023-10-23
Implement `interval` for manual push jobs

dsh2dsh:
Here is an example of configuration:
```
  - name: "zroot-to-remote"
    type: "push"
    connect:
      type: "tls"
      address: "remote:8888"
    filesystems: *filesystems
    interval: "1h"
    send:
      compressed: true
    conflict_resolution:
      initial_replication: "all"
    snapshotting:
      type: "manual"
    pruning: &pruning-remote
      keep_sender:
        - type: "regex"
          regex: ".*"
      keep_receiver: *keep-receiver
```
which doesn't do snapshots and just replicates `filesystems` every hour.

### DIFF
--- a/daemon/job/active.go
+++ b/daemon/job/active.go
@@ -102,6 +102,7 @@ type modePush struct {
 	senderConfig  *endpoint.SenderConfig
 	plannerPolicy *logic.PlannerPolicy
 	snapper       snapper.Snapper
+	interval      *config.PositiveDurationOrManual
 }
 
 func (m *modePush) ConnectEndpoints(ctx context.Context, connecter transport.Connecter) {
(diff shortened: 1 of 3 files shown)


daemon/job/active.go
```diff
@@ -132,8 +133,32 @@ func (m *modePush) Type() Type { return TypePush }
...
 
-func (m *modePush) RunPeriodic(ctx context.Context, wakeUpCommon chan<- struct{}) {
-	m.snapper.Run(ctx, wakeUpCommon)
+func (m *modePush) RunPeriodic(
+	ctx context.Context, wakeUpCommon chan<- struct{},
+) {
+	if m.interval == nil {
+		m.snapper.Run(ctx, wakeUpCommon)
+		return
+	}
```
problame, line 142: If `m.interval != nil`, then the snapper doesn't run. I don't think this is acceptable. I think we should

1. either make them mutually exclusive, i.e., error out during config parsing (and assert/panic at runtime),
2. or make them merge, i.e., spawn two goroutines, one for `m.snapper.Run()` and one for the `m.interval` ticker that you added here.

problame (changes requested):
I'm fine with the basic idea in this PR, see my review comment for requested change in behavior.

Also I wonder whether we should have this as a top-level field vs put it into `jobs[].triggers` or similar, preparing the way for my long-stalled (😞) work on a generic trigger mechanism ( https://github.com/zrepl/zrepl/pull/462#issuecomment-821842693 )

daemon/job/active.go
dsh2dsh: That's not a problem, because `m.interval != nil` only for
```yaml
    snapshotting:
      type: "manual"
```
```go
	if _, ok := in.Snapshotting.Ret.(*config.SnapshottingManual); ok {
		m.interval = in.Interval
	}
```

problame: That's still not great. The current version of this PR allows the user to configure 

```
interval: 20min
snapshotting:
  type: periodic
  interval: ...
```

The `snapshotting.interval` will take effect, and the top-level `interval` will be silently ignored.

I'll figure something out and push it into this PR

PR #768, opened 2023-11-28
add missing newline for codeblock in docs/compile-from-source.rst

Raupinger:
to make it render correctly


problame (approved):
Sorry for the long delay!

PR #785, opened 2024-05-07
allow entering time zone in timestamp_format and actually see it

Mjasnik:
if one lives in UK, UTC time zone is fine, because it matches
  local time for that user, however, if user lives somewhere else
  and quickly wants to restore a file from that particular time, he
  has to go through some mental gymnastics to calculate which time
  is the right time, because everything is stored in UTC

  for scripts, that might not be a problem, however for user, he
  interprets time as his local time, not UTC

  this patch allows user to enter and actually see the actual local
  time with TZ as a snapshot name, either numeric or otherwise
  if user did not ask for time zone, he still will get the default
  time in UTC in snapshot name

  there were couple of nuances with numeric time zones, because
  they contain - or +, - is allowed character in snapshot name,
  however + is not, so just replace + with _

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -219,11 +219,12 @@ type SnapshottingEnum struct {
 }
 
 type SnapshottingPeriodic struct {
-	Type            string            `yaml:"type"`
-	Prefix          string            `yaml:"prefix"`
-	Interval        *PositiveDuration `yaml:"interval"`
-	Hooks           HookList          `yaml:"hooks,optional"`
-	TimestampFormat string            `yaml:"timestamp_format,optional,default=dense"`
+	Type              string            `yaml:"type"`
+	Prefix            string            `yaml:"prefix"`
+	Interval          *PositiveDuration `yaml:"interval"`
+	Hooks             HookList          `yaml:"hooks,optional"`
+	TimestampFormat   string            `yaml:"timestamp_format,optional,default=dense"`
+	TimestampLocation string            `yaml:"timestamp_location,optional,default=UTC"`
 }
 
 type CronSpec struct {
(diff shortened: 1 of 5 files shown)


problame (changes requested):
I think this PR constitutes a breaking change for users who have specified a custom format in their config.
E.g., a user whose `time.Local` is UTC+2 who has custom format `2006-01-02T15:04:05.000-07` will get different output with this PR than before.

Can't have that.

The only way I'll accept this feature in zrepl is via an additional optional config field `timestamp_location`.

```yaml
jobs:
- ...
  filesystems: { ... }
  snapshotting:
    ...
    timestamp_format: 2006-01-02T15:04:05.000-07
    timestamp_location: Europe/Berlin # default: UTC
```

It should be quite obvious how to add the field to the config.
Use https://pkg.go.dev/time#LoadLocation to parse it.
Optionally also provide a way to specify a fixed offset through https://pkg.go.dev/time#FixedZone.

Then, in in `plan.formatNow()`, use `time.Now().In(plan.timestamp_zone)` to convert into the user's desired time zone (see https://pkg.go.dev/time#Time.In to convert the `time)

Mjasnik:
Ok, I see your point, but ... I need a clarification.

A user, who wanted to get local TZ by specifying it, never got it, he got UTC. He might have left that format in there and you're right, the output will be different. Is that a concern, I don't know.

Anyhow, I wanted to understand, why you explicitly want user to specify exact time zone, why not take it from the system?
If you insist on additional option, I could see this as timestamp_location may have "local" or "default" setting, which means local timezone from the system and if user specifies something other, then make a snapshot name according to that.

But then why user may want to specify completely different time zone for snapshot name other than from the system or UTC, maybe because servers are in different location than user... Maybe, I don't know.

Additionally, as user may specify all sorts of things in the format, he may add something that does not conform to snapshot names in ZFS, currently there is no guarding against that. Should we leave it as is or try to remove all non-conforming characters from format? As far as I know and ZFS suggests me ```path component must only contain alphanumeric chars and any in "-_.: "``` and alphanumeric are the only ones allowed.
Your thoughts on this?

Mjasnik:
It appears that "LoadLocation" already has "Local", so that will work just fine. Ok, I will try to implement this via timestamp_location, hopefully soon.

Mjasnik:
I think I finished this as per your suggestion, please let me know whether something else needs to be adjusted.

problame:
@Mjasnik I opened a PR with a  refactored & refined version of what you proposed in this PR.

Would appreciate a review: https://github.com/zrepl/zrepl/pull/801

PR #801, opened 2024-07-13
snapshotting: ability to specify timestamp location  != UTC

problame:
This PR adds a new field optional field `timestamp_location` that allows the user to specify a timezone different than the default UTC for use in the snapshot suffix.

I took @mjasnik 's PR https://github.com/zrepl/zrepl/pull/785 and refactored+extended it as follows:
* move all formatting logic into its own package
* disallow `dense` and `human` with formats != UTC to protect users from stupidity
* document behavior more clearly
* regression test for existing users

### DIFF
--- a/daemon/snapper/timestamp_formatting/timestamp_formatting.go
+++ b/daemon/snapper/timestamp_formatting/timestamp_formatting.go
@@ -1,0 +1,92 @@
+package timestamp_formatting
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+type Formatter struct {
+	format   func(time.Time) string
+	location *time.Location
+}
+
+func New(formatString string, locationString string) (*Formatter, error) {
+	location, err := time.LoadLocation(locationString) // no shadow
+	if err != nil {
+		errors.Wrapf(err, "load location from string %q", locationString)
+	}
+	makeFormatFunc := func(formatString string) (func(time.Time) string, error) {
+		if strings.Contains(formatString, "+") {
+			return nil, fmt.Errorf("character '+' is not allowed in ZFS snapshot names")
+		}
+		return func(t time.Time) string {
+			res := t.Format(formatString)
+			// if the formatString contains a time zone specifier
+			// and the location would result in a positive offset to UTC
+			// then the result of t.Format would contain a '+' sign.
+			if isLocationPositiveOffsetToUTC(location) {
+				// the only source of `+` can be the positive time zone offset because we disallowed `+` as a character in the format string
+				res = strings.Replace(res, "+", "_", 1)
+			}
+			if strings.Contains(res, "+") {
+				panic(fmt.Sprintf("format produced a string containing illegal character '+' that wasn't the expected case of positive time zone offset: format=%q location=%q unix=%q result=%q", formatString, location, t.Unix(), res))
+			}
+			return res
+		}, nil
+	}
+	var formatFunc func(time.Time) string
+	mustUseUtcError := func() error {
+		return fmt.Errorf("format string requires UTC location")
+	}
+	switch strings.ToLower(formatString) {
+	case "dense":
+		if location != time.UTC {
+			err = mustUseUtcError()
+		} else {
+			formatFunc, err = makeFormatFunc("20060102_150405_000")
+		}
+	case "human":
+		if location != time.UTC {
+			err = mustUseUtcError()
+		} else {
+			formatFunc, err = makeFormatFunc("2006-01-02_15:04:05")
+		}
+	case "iso-8601":
+		formatFunc, err = makeFormatFunc("2006-01-02T15:04:05.000Z0700")
+	case "unix-seconds":
+		if location != time.UTC {
+			// Technically not required because unix time is by definition in UTC
+			// but let's make that clear to confused users...
+			err = mustUseUtcError()
+		} else {
+			formatFunc = func(t time.Time) string {
+				return strconv.FormatInt(t.Unix(), 10)
+			}
+		}
+	default:
+		formatFunc, err = makeFormatFunc(formatString)
+	}
+	if err != nil {
+		return nil, errors.Wrapf(err, "invalid format string %q or location %q", formatString, locationString)
+	}
+	return &Formatter{
+		format:   formatFunc,
+		location: location,
+	}, nil
+}
+
+func isLocationPositiveOffsetToUTC(location *time.Location) bool {
+	_, offsetSeconds := time.Now().In(location).Zone()
+	return offsetSeconds > 0
+}
+
+func (f *Formatter) Format(t time.Time) string {
+	if f.location != nil {
+		t = t.In(f.location)
+	}
+	return f.format(t)
+}
--- a/config/config.go
+++ b/config/config.go
@@ -220,11 +220,16 @@ type SnapshottingEnum struct {
 }
 
 type SnapshottingPeriodic struct {
-	Type            string            `yaml:"type"`
-	Prefix          string            `yaml:"prefix"`
-	Interval        *PositiveDuration `yaml:"interval"`
-	Hooks           HookList          `yaml:"hooks,optional"`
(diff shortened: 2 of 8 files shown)


problame:
@mjasnik I would appreciate a review. You got an invite as a read-only collaborator. IIUC that should be sufficient for posting reviews.

daemon/snapper/timestamp_formatting/timestamp_formatting.go
```diff
@@ -0,0 +1,92 @@
...
+			// then the result of t.Format would contain a '+' sign.
+			if isLocationPositiveOffsetToUTC(location) {
+				// the only source of `+` can be the positive time zone offset because we disallowed `+` as a character in the format string
+				res = strings.Replace(res, "+", "_", 1)
+			}
+			if strings.Contains(res, "+") {
+				panic(fmt.Sprintf("format produced a string containing illegal character '+' that wasn't the expected case of positive time zone offset: format=%q location=%q unix=%q result=%q", formatString, location, t.Unix(), res))
+			}
+			return res
+		}, nil
```
problame, line 39: This is different than [in your PR](https://github.com/zrepl/zrepl/pull/785/files#diff-e0504de9a4bc7760b147f94c5c1b2d80e5cb699736be8c51c01acf7edd7d0d64R67-R74).

I think it's more general, but maybe I'm missing an edge case. Would appreciate close inspection of this, esp wrt regressions.

daemon/snapper/timestamp_formatting/timestamp_formatting.go
```diff
@@ -0,0 +1,92 @@
...
+	location *time.Location
+}
+
+func New(formatString string, locationString string) (*Formatter, error) {
+	location, err := time.LoadLocation(locationString) // no shadow
+	if err != nil {
+		return nil, errors.Wrapf(err, "load location from string %q", locationString)
+	}
+	makeFormatFunc := func(formatString string) (func(time.Time) string, error) {
+		if strings.Contains(formatString, "+") {
```
Mjasnik, line 23: User can enter all sorts of things in the format mask, like "\/|$#", etc. which actually work using format, I don't think guarding against "+" is exactly necessary.

daemon/snapper/timestamp_formatting/timestamp_formatting.go
```diff
@@ -0,0 +1,92 @@
...
+			return nil, fmt.Errorf("character '+' is not allowed in ZFS snapshot names")
+		}
+		return func(t time.Time) string {
+			res := t.Format(formatString)
+			// if the formatString contains a time zone specifier
+			// and the location would result in a positive offset to UTC
+			// then the result of t.Format would contain a '+' sign.
+			if isLocationPositiveOffsetToUTC(location) {
```
Mjasnik, line 31: If format against "all sorts of things" (check prev comment) is not guarded, resulting snapshot name can contain prohibited characters, so the only source of errors is not really an offset.

I would still use regexp as I used in my PR, maybe print a warning that there were invalid characters.

daemon/snapper/timestamp_formatting/timestamp_formatting.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+			return res
+		}, nil
+	}
+	var formatFunc func(time.Time) string
+	mustUseUtcError := func() error {
+		return fmt.Errorf("format string requires UTC location")
+	}
+	switch strings.ToLower(formatString) {
+	case "dense":
```
Mjasnik, line 46: In this case, I would actually print a warning maybe (or maybe not) that location is not UTC and force it to be UTC in code, and not give an error, because predefined format masks require implicit use of UTC.

problame: > and not give an error,

The error will force the user to recognize that their configuration is incorrect.
The default `timestamp_location` will be UTC, so,

```
timestamp_format: dense
```

will continue to work without error.

Only if they explicitly specify something wrong, like,

```yaml
timestamp_format: dense
timestamp_location: Europe/Berlin
```

will there be an error.

problame:
Also, have you tried this PR for your use case?

Mjasnik:
> Regarding the disallowed characters in snapshot names,
> 
> > like "/|$#",
> 
> maybe reimplementing the [`snapshot_namecheck`](https://github.com/openzfs/zfs/blob/1d3ba0bf01020f5459b1c28db3979129088924c0/module/zcommon/zfs_namecheck.c#L334-L351) function in Go would be the better way.
> 
> Then, we could fail config parsing if the timestamp doesn't pass the namecheck, which is better than a late runtime failure.
> 
> This works because luckily a standard go time format string doesn't contain any prohibited characters.
> 
> We should run the same namecheck function over the `prefix: ` field as well, if we add this.
> 
> WDYT?

Well, I think the best way to check configuration is when zrepl starts and not when it is about to make a snapshot. That's it. Early warning at the start is always better, than dealing with it later.

Mjasnik:
> Also, have you tried this PR for your use case?

No, I am running my PR on all machines I need. This is partly because this needs to be stable and should not fail, and I tested my
 PR. I am waiting when something stable comes out of this PR and then I'll test it :)

problame:
It seems you're implying that the current state of this PR would fail more at runtime than your original PR? If so, would love for a pointer to that code, I think they're equivalent wrt runtime failure risk.

I think In going to implement the snapshot_namecheck approach, it'll be an improvement over not just what's in this PR but also `master`.

Mjasnik:
> It seems you're implying that the current state of this PR would fail more at runtime than your original PR? If so, would love for a pointer to that code, I think they're equivalent wrt runtime failure risk.
> 
> I think In going to implement the snapshot_namecheck approach, it'll be an improvement over not just what's in this PR but also `master`.

No, I am not implying much here. I have a stable version which works, PR was made 2 months ago and only now it started moving. When this will be finalized, I'm going to try it.

daemon/snapper/snapname/snapname.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	formatter := &Formatter{
+		prefix:    prefix,
+		timestamp: timestamp,
+	}
+	// Best-effort check to detect whether the result would be an invalid name early.
+	testFormat := "some/dataset@" + formatter.Format(time.Now())
+	if err := zfs.EntityNamecheck(testFormat, zfs.EntityTypeSnapshot); err != nil {
+		return nil, errors.Wrapf(err, "prefix or timestamp format result in invalid snapshot name such as %q", testFormat)
+	}
```
problame, line 30: This check isn't 100% correct / selft-contained.

For example, imagine out `timestamp.Formatter` didn't have the special handling for `+` timezone offsets.

If you're in the UK, the `+` appears only in the summer (BST is UTC+1)

=> The check would only fail in the summer. But we'd want it to fail all the time.

daemon/snapper/timestamp_formatting/timestamp_formatting.go
problame: >  so the only source of errors is not really an offset.

Addressed with recent pushes ; we now do the same check that ZFS does.

problame: > I would still use regexp as I used in my PR, maybe print a warning that there were invalid characters.

We now detect this shortly after config parsing during startup, and will fail to start the daemon.

daemon/snapper/timestamp_formatting/timestamp_formatting.go
problame: Added a comment to explain why we're checking for `+` specifically.

Also, with recent pushes, there's now a config-parsing-time check for all invalid characters: https://github.com/zrepl/zrepl/blob/a8ece782ef0607257b8234dd90fd24f8761b4119/daemon/snapper/snapname/snapname.go#L27-L42

problame:
@Mjasnik I added the config-parsing-time check for all invalid characters.

I think this is good to merge, would be happy if you could test it.

Mjasnik:
I built zrepl from master + this MR and the result looks good to me, name check works and timezone names work as expected in my usecase.
I would say that this MR is ready.

problame:
Thanks for testing!

PR #806, opened 2024-07-26
Remove dead Bountysource link

lpulley:
Bountysource is dead.

https://www.theblockchain-group.com/wp-content/uploads/2023/11/TBG-CP17112023.pdf


problame (approved):
Thanks for this!

PR #810, opened 2024-07-27
Add an attention for `not_replicated` in doc

wxiaoguang:
I am not native speaker, feel free to edit the content directly.

Thank you.

(or if it is not suitable, feel free to close it)


problame:
Thanks for your contribution! I changed the wording a bit, but it's a legitimate issue.
The quickstart guides' examples still include it, so, presumably, many users use it by default.
I considered removing it, but, for the typical server + "scheduled backup" use case, I think it's better to have it.

PR #815, opened 2024-08-10
Serve Prometheus metrics over (m)TLS

frebib:
Add a TLSConfig struct/library to configure a simple TLS/HTTPS server with most basic toggles that most users would want. This is based on the structure that the prometheus/exporter-toolkit web-config uses.

Notably this does not make use of either of the two existing TLS configuration structures that exist as both are specific in their usage and neither are suitable for generally configuring a TLS server.

### DIFF
--- a/config/tls_test.go
+++ b/config/tls_test.go
@@ -1,0 +1,179 @@
+package config
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"os"
+	"path"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/zrepl/yaml-config"
+)
+
+func TestTLSConfig(t *testing.T) {
+	input := `
+cert: path/to/cert.pem
+key: path/to/key.pem
+client_ca: path/to/ca.pem
+client_auth_type: RequireAndVerifyClientCert
+min_version: TLS1.0
+max_version: TLS1.3
+cipher_suites:
+  # TLS 1.2 cipher suites.
+  - TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA
+  - TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA
+  - TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
+  - TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA
+  - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+  - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
+  - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+  - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+  - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
+  - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
+  # TLS 1.3 cipher suites.
+  - TLS_AES_128_GCM_SHA256
+  - TLS_AES_256_GCM_SHA384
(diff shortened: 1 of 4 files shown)


config/tls_test.go
```diff
@@ -0,0 +1,179 @@
...
+	var actual TLSConfig
+	err := yaml.Unmarshal([]byte(input), &actual)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !reflect.DeepEqual(actual, parsed) {
+		t.Fatal("parsed TLS config doesn't match expected")
+	}
+
+	workdir := t.TempDir()
```
frebib, line 81: This is a Golang 1.15 feature, but I notice that this project is still stuck on the ancient Go 1.12 (not sure why, but I tried my best within the constraints).
I assumed that this was fine to break that constraint given that this is just test code

problame:
Sorry for the delay in reviewing this PR.

>  This is based on the structure that the prometheus/exporter-toolkit web-config uses.

Why not pull in the `prometheus/exporter-toolkit`s `TLSConfig` instead of rolling our own here?

I agree that the functionality added in this PR is useful (exporting zrepl prometheus metrics over TLS) but I don't want to maintain a `YAML <=> tls.Config` mapping.

> I notice that this project is still stuck on the ancient Go 1.12

Ah, I see, the `go 1.12` line in the `go.mod` file.
Bump it as you need to get this PR done.
zrepl is behind on bumping its Go dependencies, so, possibly there might be trouble upgrading to >= `go 1.21` , which is when the Go toolchain starts enforcing

> At go 1.21 or higher:
> 
> * The go line declares a required minimum version of Go to use with this module.
> * The go line must be greater than or equal to the go line of all dependencies.
> * The go command no longer attempts to maintain compatibility with the previous older version of Go.
> * The go command is more careful about keeping checksums of go.mod files in the go.sum file.
> 
> [https://go.dev/ref/mod#go-mod-file-go]

frebib:
I'm happy to rework this and pull in `prometheus/exporter-toolkit`, but it's going to mean a bump of minimum golang version to 1.22 https://github.com/prometheus/exporter-toolkit/blob/master/go.mod#L3

If you're happy with that, then I'll change it up- let me know

problame:
Hmm, I guess I'll have to bump a lot of dependencies first.

I will do that in a separate PR, but to unblock you, just do whatever bumping you need to do to pull in the exporter-toolkit.

I'll rebase your PR when I'm done.

problame:
For the record, WIP branch with dependency updates here: https://github.com/zrepl/zrepl/pull/819

frebib:
Thanks for the update. Yes, I plan to pick this up at some point when time allows

frebib:
Dropping the TLSConfig struct made this patch a lot smaller. So small in fact that I'm having a hard time believing it 😆  Testing it now

frebib:
Seems like it's gonna need some more work. Even ignoring the subtle differences I introduced, it doesn't seem to accept my config

problame:
Transitioning this back to draft state, please reopen and ping me once ready.

frebib:
Looks like the problem is zrepl's custom YAML decoder: https://github.com/go-yaml/yaml/commit/3fc339b5b0e8c5f2d68bf82a66cd2187c0787711
The upstream prometheus/exporter-toolkit config structs are written to work with go-yaml/yaml and not forks that change unmarshaling behaviour.

We could revert this patch to the initial implementation and carry the burden of having to re-implement `TLSConfig`

  > Why not pull in the `prometheus/exporter-toolkit`s `TLSConfig` instead of rolling our own here?

  I guess we found a reason why 😁 

Alternatively we could change zrepl using the forked yaml implementation, or maybe revert `optional=true` to `required=false`  to be more compatible with other external structs that expect to be parsed from yaml.

As-is, I don't see any other ways forward. Let me know what you want to do @problame

problame:
Apologies for letting this sit for so long, life got in the way.

> edit: one more option I think might be viable (and maybe a reasonable middle-ground) is to reimplement the TLSConfig struct _just_ for unmarshaling, then copy all fields into the `prometheus/exporter-toolkit` struct for use in the application?

This sounds acceptable, let me know if you want to pick this up, or close this PR if you (understandably) lost interest in contributing.

Issue #708, opened 2023-05-23
mutiple configuration files


pterrien:
I find `zrepl` really great, but I must admit I miss the possibility to split the configuration into multiple files, as currently I find the single `zrepl.yaml` sometimes difficult to maintain.

Here is how I would see it:
`<config_folder>/zrepl.yaml`  must have the `global` section, and optionally a `jobs` one
`<config_folder>/jobs.d/*.yaml`: each file contains a list of jobs

The approach being a simple merge of all `jobs.d/*yaml` files into the `jobs:` section.
Of course, we would have to handle the case when a job has already been defined, in the main config or another `jobs.d/` file, but I believe a simple rule like "the latter / former takes precedence" would be perfectly acceptable (avoiding the difficulty to merge job definitions, which would be a nightmare).

We could also imagine a new command `zrepl config`, that would display the computed configuration.

Do you think this would be useful, and not to difficult to implement? Or are there some caveats I'm not aware of?

Thanks.

problame:
> Do you think this would be useful,

Maybe, I personally don't need it, and for complex setups, I wonder whether using a tool like Ansible to create the zrepl.yml would be easier anyways (cc @InsanePrawn )

> and not to difficult to implement?

Quite easy to implement, I think.
I think instead of having a precedence rule, I'd just enforce a 'no overlap' rule, i.e., there must be no duplicate job names.

---

I think syntactically, you could make it backwards compatible with

```yaml
global:
  ...

# either
jobs:
  - ...
  - ...
# or
jobs:
  include: ./jobs.d
```

Now that I'm typing this out, the next question is about relative vs abs paths vs both in the `include`.
To what base would we evaluate the relative path to?
zrepl has the path through which it found the config file handy, so, I guess we could use the zrepl.yml config file itself as a base?


---

generally I'm feeling not very enthusiastic about this

pterrien:
Thanks for your answer.

> Maybe, I personally don't need it, and for complex setups, I wonder whether using a tool like Ansible to create the zrepl.yml would be easier anyways (cc @InsanePrawn )

Indeed, I haven't thought about Ansible, my plan B was to write a simple merge script in Python or so.

> I think instead of having a precedence rule, I'd just enforce a 'no overlap' rule, i.e., there must be no duplicate job names.

Yes, probably better this way, in any case no having to merge job with same is the most important, otherwise it would make things far more complex.

> Now that I'm typing this out, the next question is about relative vs abs paths vs both in the include.
[...]

PR #856, opened 2024-12-01
Added distributed job YAML file support

zeyadtamimi:
**Note: This is a WIP PR that presently only adds the documentation for the features so that proposal can be discussed as per #708 **

Background
==========

While trying to use zrepl in my ansible driven home lab deployment I ran into an interesting problem. My ZFS based servers subscribe to different sometimes overlapping roles.

Example role distribution between two servers:
```
serverA:
  - common
  - web
  - file

serverB:
  - common
  - git
  - file
```

Each role wants to create and manage a ZFS dataset with its own replication / backup policies:
   - web: pool/web
   - git: pool/git
   - file pool/file

At present the creation of a ZFS dataset from each role role is somewhat very easy, so to is the creation of the basic zrepl configuration file from the "common" role.

However, when each role tries to register it's job(s) into the singular zrepl configuration files things get tricky.

I could try adding a role at the end that hardcodes the datasets that need to be backed up but that seems a bit hacky.

I could also use ansible's `lineinfile` task to try to idempotently add each dataset's snapshot jobs to the zrepl configuration files but that causes a problem:

Everytime the "common" role gets run the basic zrepl configuration file gets re-created causing the web, git, and file roles to all register changes as they have to re-insert all jobs back into the singular configuration file.

Proposed Solution
=================

The proposed solution is to allow for the distribution of zrepl job definition between multiple different YAML files that can be included from the main zrepl configuration files.

```
global: ...
jobs:
  include: jobs.d
```

This directive would be only acceptable in the main configuration file and is mutually exclusive with any other job definitions in the file.

To keep things lean there will be no conflict resolution provided to users, job names must be unique across all included job YAML files.

With this feature, the above problem becomes much simpler:
   - Common: Sets up the global zrepl configuration and the include directive
   - web/git/file: Each manage their own datasets and create their jobs.d/web.yml, jobs.d/git.yml, and jobs.d/file.yml.

### DIFF
--- a/docs/configuration/misc.rst
+++ b/docs/configuration/misc.rst
@@ -44,3 +44,54 @@ Interval & duration fields in job definitions, pruning configurations, etc. must
 
     var durationStringRegex *regexp.Regexp = regexp.MustCompile(`^\s*(\d+)\s*(s|m|h|d|w)\s*$`)
     // s = second, m = minute, h = hour, d = day, w = week (7 days)
+
+
+.. _conf-include-directories:
+
+Including Configuration Directories
+-----------------------------------
+
+It is possible to distribute zrepl job definitions over multiple YAML files. This is
+achieved by using the `include` key in the configuration file.
+
+The directive is mutually exclusive with any other jobs definition and can only exist
+in the configuration file:
+
+::
+   global: ...
+
+   jobs:
+     include: jobs.d
+
+.. NOTE::
+   The included directory path is treated as absolute when starting with `/` else it
+   is treated as a relative path from the directory of the loaded configuration file.
+
+Included YAML job files must end with the `.yml` extension and can only contain
+contain the `jobs` key. Additionally, job names must be unique across all job YAML
+files.
+
+Examples
+^^^^^^^^
+
+::
+  > /etc/zrepl/zrepl.yml
+
+   jobs:
+     include: jobs.d
+
+  > /etc/zrepl/zrepl.d/MyDataSet.yml
+   jobs:
+   - name: snapjob
+     type: snap
+     filesystems: {
+       "MyPool/MyDataset<": true,
+     }
+     snapshotting:
+       type: periodic
+       interval: 2m
+       prefix: zrepl_snapjob_
+     pruning:
+       keep:
+         - type: last_n
+           count: 60


docs/configuration/misc.rst
```diff
@@ -44,3 +44,54 @@ Interval & duration fields in job definitions, pruning configurations, etc. must
...
+.. _conf-include-directories:
+
+Including Configuration Directories
+-----------------------------------
+
+It is possible to distribute zrepl job definitions over multiple YAML files. This is
+achieved by using the `include` key in the configuration file.
+
+The directive is mutually exclusive with any other jobs definition and can only exist
+in the configuration file:
```
InsanePrawn, line 58:
```suggestion
in the main configuration file:
```

docs/configuration/misc.rst
```diff
@@ -44,3 +44,54 @@ Interval & duration fields in job definitions, pruning configurations, etc. must
...
     // s = second, m = minute, h = hour, d = day, w = week (7 days)
+
+
+.. _conf-include-directories:
+
+Including Configuration Directories
+-----------------------------------
+
+It is possible to distribute zrepl job definitions over multiple YAML files. This is
+achieved by using the `include` key in the configuration file.
```
InsanePrawn, line 55:
```suggestion
achieved by using the `include` key in the main configuration file.
```

docs/configuration/misc.rst
```diff
@@ -44,3 +44,54 @@ Interval & duration fields in job definitions, pruning configurations, etc. must
...
+Examples
+^^^^^^^^
+
+::
+  > /etc/zrepl/zrepl.yml
+
+   jobs:
+     include: jobs.d
+
+  > /etc/zrepl/zrepl.d/MyDataSet.yml
```
InsanePrawn, line 83:
```suggestion
  > /etc/zrepl/jobs.d/MyDataSet.yml
```

InsanePrawn (commented):
Found a typo and suggested some improvements for clarity.

Generally a good proposal IMO.

I think the format could use some discussion.
I've got some questions:
1. Why put the includes under `jobs:`?
2. Why only allow a single include?
3. Should we allow the user to specify a single file (`jobs.d/` vs `jobs.d/specificfile.yml`)?

Tl;dr: is there a reason why it's not
```yaml
includes:
  - jobs.d/
  - /etc/zrepl/this_one_really_weird_file.yml
  ```
  
  wondering what @problame thinks about either approaches

**Comment by zeyadtamimi** [comment]
Thanks for the review

>     1. Why put the includes under `jobs:`?
>        Oh, I see you copied it from the issue. In that case @problame: I suspect it's because you wanted to emphasize the fact that only job definitions should go in other config files, but why not raise an error for now on non-job keys in extension files and allow that possibility later if we find it beneficial?

Yeah I tried to preserve some of the suggestions in the original issue. I personally agree with you that having a separate key makes more sense. Though I would prefer something more explicit like `job_includes`.

>     2. Why only allow a single include?
No particular reason other than to simplify the potential implementation ( e.g. Zrepl does not need to handle the case where the file `a/b.yml` is included along with the directory `a/`

>     3. Should we allow the user to specify a single file (`jobs.d/` vs `jobs.d/specificfile.yml`)?

I am not opposed to the idea, It just does not really benefit the proposed use case where the entity writing the configuration file is different than the one writing the jobs file.

Aside from my preference for `job_includes` I don't really see an issue with your suggested improvements to the PR.

Let me update the request.

File: docs/configuration/misc.rst [hunk 4]
```diff
@@ -44,3 +44,54 @@ Interval & duration fields in job definitions, pruning configurations, etc. must
...
+files.
+
+Examples
+^^^^^^^^
+
+::
+  > /etc/zrepl/zrepl.yml
+
+   jobs:
+     include: jobs.d
```
problame, line 81: So, it's all or nothing? Either I specify a list of jobs in the main config file or I specify a path where the files are located?

docs/configuration/misc.rst
```diff
@@ -44,3 +44,54 @@ Interval & duration fields in job definitions, pruning configurations, etc. must
...
+
+::
+   global: ...
+
+   jobs:
+     include: jobs.d
+
+.. NOTE::
+   The included directory path is treated as absolute when starting with `/` else it
+   is treated as a relative path from the directory of the loaded configuration file.
```
problame, line 68: Regarding relative paths, I think it'd be more precise to state that the relative path is added to the `filepath.Dir()` of the main config file path used by the daemon. With a link to the list of default config locations [here](https://zrepl.github.io/configuration/overview.html#overview-terminology).

docs/configuration/misc.rst
problame: Also, on the relative paths, I suggest to use `./jobs.d` so that it's clear this is a relative path and not some string.

problame (commented):
Thanks for going docs first!

I like the proposal with the top-level `include` + only allowing `jobs` as a top-level key in the included files.
Keeps things open for extension later on.

Also, it makes it removes ambiguity with regards to how one parses `jobs`.

---

Please digest the feedback into an update to this PR (just new commits, no force pushes please) and ask for re-review.

I think after that implementation can start.

docs/configuration/misc.rst
zeyadtamimi: Fixed

zeyadtamimi:
Thanks for the review.

> I like the proposal with the top-level `include` + only allowing `jobs` as a top-level key in the included files. Keeps things open for extension later on.

In that case should we still have `jobs` and the include token be mutually exclusive in the main configuration file?

docs/configuration/misc.rst
zeyadtamimi: Actually thinking more about this if we have the include as a separate top level key we don't need this mutual exclusivity.

Will update the PR.

zeyadtamimi:
@problame I modified the proposal based on your and @InsanePrawn's comments.

Please take a look and let me know what you think of the new usage semantics. I also opted to move the documentation into a new file rather than cramming it into `misc`.

zeyadtamimi:
@problame gentle bump on the review

zeyadtamimi:
@problame I have finished an initial implementation and added some basic tests.

Are you happy with the usage semantics? If so I will push the initial implementation into the PR.

InsanePrawn:
> I have finished an initial implementation and added some basic tests.

Very nice! Please do push your implementation, we can always change the code.

> In that case should we still have jobs and the include token be mutually exclusive in the main configuration file?

IMO no, we're gonna need to detect job name conflicts between drop-in files anyway, no need to treat the main config file differently IMO. (Am i overlooking anything?)

zeyadtamimi:
Hey @InsanePrawn / @problame,

Sorry this got dropped from my radar for a while since I was struggling with getting the lint job to pass. Finally got back to it. let me know what you think

internal/config/config.go
```diff
@@ -679,14 +683,79 @@ func ParseConfig(path string) (i *Config, err error) {
...
+}
+
+func ExpandConfigInclude(configPath string, config *Config) (err error) {
+	if config == nil {
+		return nil
+	}
+
+	var includeConfigPaths []string
+	for _, path := range config.Include {
+		if configPath[0] != '/' {
```
problame:
```suggestion
		if !pathpkg.IsAbs(configPath) {
```

PR #858, opened 2024-12-09
Fix link to hooks sample template

JMoVS:


problame:
Thanks for being proactive about this. However, the issue really is that zrepl.github.io hasn't been deployed since the renaming happened.

Will do that shortly.

PR #859, opened 2024-12-11
properties -> send_properties and add nfs/smb note for property replication

Malvineous:


problame (approved):

problame:
Thanks for this contribution!

PR #864, opened 2025-01-06
more Grafana changes (for https://github.com/zrepl/zrepl/pull/847)

problame:
This PR stacks on top of PR https://github.com/zrepl/zrepl/pull/847 

I couldn't push there because [you have disabled maintainer changes](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

I added a new metric `zrepl_daemon_job_up` for the job name.

I think it's useful to have that metric, but, obviously old scrapes don't have it, so, the dashboard will likely show "No Data". 

So, still considering whether we can do without that metric.

image

PR #865, opened 2025-01-06
add metric for exposing job names & tracking `up`-ness

problame:
Spin-off from https://github.com/zrepl/zrepl/pull/864

Metric name is `zrepl_daemon_job_up`


problame:
Wasn't necessary, let's throw this code away.

PR #877, opened 2025-03-19
Docs: add san subject-alt-name to the easyrsa script

Schueni1:
The current easyrsa script is not compatible with zrepl 0.3.0 and newer because of the missing san entry.


problame:
Hm, are you using a different easyrsa version than the (admittedly) old one in the script? 
I get the same CN and SAN with old and new script on my machine.

PR #932, opened 2026-02-27
rpm: Use %config(noreplace) for the config file

chenxiaolong:
Currently, when upgrading the RPM, if the upstream `zrepl.yml` is changed in the update, the user's `zrepl.yml` will be renamed to `zrepl.yml.rpmsave` and the new `zrepl.yml` is installed in its place.

This commit changes the behavior so that the user's config remains in place and the new upstream config is installed to `zrepl.yml.rpmnew` instead.

---

This was never a problem before because `packaging/systemd-default-zrepl.yml` was the same since the initial release. RPM leaves edited config files alone when the upstream config file doesn't change. However, v0.7.0 changed the file with: 9c63736489569adf07537b9b01dc5e84f23f9651. So with this update, RPM detects an upstream config change during the update and moves the user's config according to `%config`'s rules (https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html). `%config(noreplace)` prevents this from happening.


problame:
This is unfortunate, thanks for fixing this. Going to figure out whether I'll do a full new release or just an RPM release.

